PR #4, opened 2022-12-15
HIPP-199: remove identity action to skip auth

laurahk:
This is not the entire ticket, but if merged now will unblock other tickets in progress (hence the name of the branch isn't reflective of the work in the PR). 

There's lots of unused code (more than just `IdentifierAction.scala`), but I'm tempted to leave everything as is until we complete the login work incase it's useful - then do a tidy up PR. 

2 more PRs to follow: 

1. add the login link to the landing page
2. remove unused import warnings in tests

### DIFF
--- a/app/controllers/IndexController.scala
+++ b/app/controllers/IndexController.scala
@@ -16,20 +16,19 @@
 
 package controllers
 
-import controllers.actions.IdentifierAction
-import javax.inject.Inject
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import views.html.IndexView
 
+import javax.inject.Inject
+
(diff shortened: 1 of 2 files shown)


app/controllers/IndexController.scala
```diff
@@ -16,20 +16,19 @@
...
 package controllers
 
-import controllers.actions.IdentifierAction
-import javax.inject.Inject
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendBaseController
 import views.html.IndexView
 
+import javax.inject.Inject
```
laurahk, line 24: I use IntelliJ and do 'optimise imports' which is why this moved. We need to agree on a standard (all use default intelliJ) or introduce scalafmt or something. To be discussed in the new year.

PaulCDurham (approved):

PR #9, opened 2023-01-04
HIPP-39: Add Application Name page

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-39

Use scaffold to create page for Application Name.

### DIFF
--- a/app/viewmodels/checkAnswers/ApplicationNameSummary.scala
+++ b/app/viewmodels/checkAnswers/ApplicationNameSummary.scala
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package viewmodels.checkAnswers
+
+import controllers.routes
+import models.{CheckMode, UserAnswers}
+import pages.ApplicationNamePage
+import play.api.i18n.Messages
+import play.twirl.api.HtmlFormat
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryListRow
+import viewmodels.govuk.summarylist._
+import viewmodels.implicits._
+
+object ApplicationNameSummary  {
+
+  def row(answers: UserAnswers)(implicit messages: Messages): Option[SummaryListRow] =
+    answers.get(ApplicationNamePage).map {
+      answer =>
+
+        SummaryListRowViewModel(
+          key     = "applicationName.checkYourAnswersLabel",
+          value   = ValueViewModel(HtmlFormat.escape(answer).toString),
+          actions = Seq(
+            ActionItemViewModel("site.change", routes.ApplicationNameController.onPageLoad(CheckMode).url)
+              .withVisuallyHiddenText(messages("applicationName.change.hidden"))
+          )
+        )
+    }
+}
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -118,5 +118,5 @@ tracking-consent-frontend {
 }
 
 features {
-  welsh-translation: true
+  welsh-translation: false
 }
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -35,8 +35,8 @@ error.number = Please enter a valid number
 error.required = Please enter a value
 error.summary.title = There is a problem
 
-index.title = The API Hub
-index.heading = Your HIP Dashboard
+index.title = Applications
+index.heading = Applications
 index.registerApplication = Register an application
 
 checkYourAnswers.title = Check Your Answers
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -56,8 +56,9 @@ signedOut.guidance = We did not save your answers.
 unauthorised.title = You can’t access this service with this account
 unauthorised.heading = You can’t access this service with this account
 
-applicationName.title = applicationName
-applicationName.heading = applicationName
+applicationName.title = Application Name
+applicationName.heading = What is the name of your application?
+applicationName.hint = This name will be used to identify your application to you and your team
 applicationName.checkYourAnswersLabel = applicationName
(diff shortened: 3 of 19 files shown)


conf/messages.en
```diff
@@ -35,9 +35,9 @@ error.number = Please enter a valid number
 error.required = Please enter a value
 error.summary.title = There is a problem
 
-index.title = api-hub-frontend
-index.heading = api-hub-frontend
-index.guidance = Welcome to your new frontend. Please see the README file for a guide to getting started.
+index.title = The API Hub
+index.heading = Your HIP Dashboard
```
laurahk, line 39: Please can we change this to "Your Dashboard"

test/controllers/IndexControllerSpec.scala
```diff
@@ -41,5 +49,37 @@ class IndexControllerSpec extends SpecBase {
         contentAsString(result) mustEqual view()(request, messages(application)).toString
       }
     }
+
+    "must initiate User Answers and redirect to the Application Name page for a POST" in {
+
+      val mockSessionRepository = mock[SessionRepository]
+
+      when(mockSessionRepository.set(any())) thenReturn Future.successful(true)
```
laurahk, line 57: I tried to understand what returning `true` from the `set` method was which I believe just means it worked and didn't error? Because there doesn't seem to be a scenario where it returns `false`. Is that correct?

test/controllers/IndexControllerSpec.scala
```diff
@@ -41,5 +49,37 @@ class IndexControllerSpec extends SpecBase {
...
+        val result = route(application, request).value
+
+        status(result) mustEqual SEE_OTHER
+        redirectLocation(result).value mustEqual routes.ApplicationNameController.onPageLoad(NormalMode).url
+
+        val userAnswersCaptor: ArgumentCaptor[UserAnswers] = ArgumentCaptor.forClass(classOf[UserAnswers])
+        verify(mockSessionRepository).set(userAnswersCaptor.capture())
+
+        val actualUserAnswers = userAnswersCaptor.getValue
+        actualUserAnswers.id mustBe "id"
```
laurahk, line 78: The `Json.obj()` is a defaulted value, but where does the string "id" come from?

app/viewmodels/checkAnswers/ApplicationNameSummary.scala
```diff
@@ -0,0 +1,43 @@
...
+import controllers.routes
+import models.{CheckMode, UserAnswers}
+import pages.ApplicationNamePage
+import play.api.i18n.Messages
+import play.twirl.api.HtmlFormat
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryListRow
+import viewmodels.govuk.summarylist._
+import viewmodels.implicits._
+
+object ApplicationNameSummary  {
```
laurahk, line 28: This was all generated?

test/pages/ApplicationNamePageSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+package pages
+
+import pages.behaviours.PageBehaviours
+
+
+class ApplicationNamePageSpec extends PageBehaviours {
+
+  "ApplicationNamePage" - {
+
+    beRetrievable[String](ApplicationNamePage)
```
laurahk, line 26: What are these tests doing?

laurahk (changes requested):
All looks fab 👍 
Just the title changes to be made.

test/controllers/IndexControllerSpec.scala
PaulCDurham: Looking in the code for SessionRepository it always returns true. An error should result in a runtime exception when the future is resolved.

conf/application.conf
```diff
@@ -118,5 +118,5 @@ tracking-consent-frontend {
 }
 
 features {
-  welsh-translation: true
+  welsh-translation: false
```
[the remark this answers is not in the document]
PaulCDurham: It's configurable so I would prefer to simply configure it out.

test/controllers/IndexControllerSpec.scala
PaulCDurham: There is a test double called `FakeIdentifierAction` that simply hard codes the User Id as "id". You can look at SpecBase to see what doubles are injected by default.

app/viewmodels/checkAnswers/ApplicationNameSummary.scala
PaulCDurham: Yup. It'll be very useful in the ticket to get Check Your Answers working. This basically puts the title/value/change button in the answers summary section.

test/pages/ApplicationNamePageSpec.scala
PaulCDurham: They are called behaviours, tests that can be reused within other tests. A useful part of ScalaTest. If you look within `PageBehaviours` you can see the full spec of each behaviour.

conf/messages.en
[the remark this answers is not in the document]
PaulCDurham: Done

conf/messages.en
```diff
@@ -35,9 +35,9 @@ error.number = Please enter a valid number
 error.required = Please enter a value
 error.summary.title = There is a problem
 
-index.title = api-hub-frontend
-index.heading = api-hub-frontend
-index.guidance = Welcome to your new frontend. Please see the README file for a guide to getting started.
+index.title = The API Hub
```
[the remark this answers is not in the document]
PaulCDurham: Set to "Applications"

conf/messages.en
```diff
@@ -35,9 +35,9 @@ error.number = Please enter a valid number
 error.required = Please enter a value
 error.summary.title = There is a problem
 
-index.title = api-hub-frontend
-index.heading = api-hub-frontend
-index.guidance = Welcome to your new frontend. Please see the README file for a guide to getting started.
```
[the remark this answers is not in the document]
PaulCDurham: It might take a while to work out what is redundant. For example there are messages for yes/no/dates that we're not using but we haven't created any yes/no or date pages yet.

conf/application.conf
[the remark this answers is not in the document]
laurahk: Good point about the updates!

test/controllers/IndexControllerSpec.scala
laurahk: Only thought would be to name the boolean, something like "success" but it's not important.

laurahk (approved):

test/controllers/IndexControllerSpec.scala
PaulCDurham: I found a constant in SpecBase that makes it a bit more understandable than the hard-coded "id" so used that.

laurahk (approved):

PR #10, opened 2023-01-09
HIPP-39: Stride Login

PaulCDurham:
Swap from using auth-login-stub to authenticating with Stride.

### DIFF
--- a/app/controllers/actions/IdentifierAction.scala
+++ b/app/controllers/actions/IdentifierAction.scala
@@ -49,8 +49,8 @@ class AuthenticatedIdentifierAction @Inject()(
 
     authorised().retrieve(Retrievals.credentials) {
       _.map {
-        credentials => block(IdentifierRequest(request, credentials.providerId))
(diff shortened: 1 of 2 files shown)


app/controllers/actions/IdentifierAction.scala
```diff
@@ -33,22 +35,25 @@ trait IdentifierAction extends ActionBuilder[IdentifierRequest, AnyContent] with
 
 class AuthenticatedIdentifierAction @Inject()(
                                                override val authConnector: AuthConnector,
-                                               config: FrontendAppConfig,
-                                               val parser: BodyParsers.Default
+                                               configuration: FrontendAppConfig,
+                                               val parser: BodyParsers.Default,
+                                               override val config: Configuration,
```
laurahk, line 40: Why are these `override`s?

laurahk: nevermind, I see now they are overriding values in `AuthRedirects`

laurahk (approved):

PR #11, opened 2023-01-09
HIPP-200: Add Application Name to Check Your Answers

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-200

### DIFF
--- a/test/controllers/CheckYourAnswersControllerSpec.scala
+++ b/test/controllers/CheckYourAnswersControllerSpec.scala
@@ -17,16 +17,23 @@
 package controllers
 
 import base.SpecBase
+import controllers.CheckYourAnswersControllerSpec.buildSummaryList
+import generators.Generators
+import models.UserAnswers
+import org.scalatestplus.scalacheck.ScalaCheckPropertyChecks
+import play.api.i18n.Messages
 import play.api.test.FakeRequest
 import play.api.test.Helpers._
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.SummaryList
+import viewmodels.checkAnswers.ApplicationNameSummary
 import viewmodels.govuk.SummaryListFluency
 import views.html.CheckYourAnswersView
 
-class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency {
+class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency with ScalaCheckPropertyChecks with Generators {
 
   "Check Your Answers Controller" - {
 
-    "must return OK and the correct view for a GET" in {
+    "must return OK and the correct view for a GET with empty user answers" in {
 
       val application = applicationBuilder(userAnswers = Some(emptyUserAnswers)).build()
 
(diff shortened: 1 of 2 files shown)


test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -36,13 +43,33 @@ class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency {
...
+        running(application) {
+          val request = FakeRequest(GET, routes.CheckYourAnswersController.onPageLoad.url)
+
+          val result = route(application, request).value
+
+          val view = application.injector.instanceOf[CheckYourAnswersView]
+          val list = buildSummaryList(userAnswers, messages(application))
+
+          status(result) mustEqual OK
+          contentAsString(result) mustEqual view(list)(request, messages(application)).toString
```
laurahk, line 67: I know these type of tests came from the scaffold, but personally I don't find these tests very clear. The comparison is so broad. 

I wonder if either replace (or have an additional test) where we could parse out the bits we are interested in. e.g. the assertion looks like actualRows equal "My-app-name" type of thing. Or actualRows.length is empty. We could still use generators in this approach.

What are your thoughts?

test/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -36,13 +43,33 @@ class CheckYourAnswersControllerSpec extends SpecBase with SummaryListFluency {
...
+        val list = buildSummaryList(emptyUserAnswers, messages(application))
 
         status(result) mustEqual OK
         contentAsString(result) mustEqual view(list)(request, messages(application)).toString
       }
     }
 
+    "must return OK and the correct view for a GET with complete user answers" in {
+
+      forAll((userAnswers: UserAnswers) => {
```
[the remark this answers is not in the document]
PaulCDurham: The previous test uses an empty user answers so covers your first point, ie do we correctly render empty user answers. This test relates specifically to complete user answers.

I suggest we address journey recovery after MVP. That covers missing user answers and `requireData`.

We also dismissed any implementation of data validation prior to MVP. We'd need to think how to resolve invalid data in the front end, ie highlight what is missing/invalid or simply start again. We may get a good steer by looking at what other services have done. I remember we had to have a good think about this on PODS.

We're probably going to come up with several areas for improvement now the service is up and running.

test/controllers/CheckYourAnswersControllerSpec.scala
PaulCDurham: Property based testing moves away from single test values and uses generators which produce multiple random/arbitrary values. I guess you're more questioning whether we like that approach which is probably something to discuss as a team.

laurahk: No, I'm board with property based testing (although I don't think every single test needs to be written this way - only when it makes sense).

I'm questioning the comparison of a stringified view against view. I don't know if you've seen the test output when it fails, but it's not particularly easy to debug. Instead of `contentAsString` we could have helper functions that extract the bits we care about. 

To me this test says "is this html I get from visiting a page the same html that is created when I pass x,y,z variables to the constructor". Rather than, "does the html I get from visiting a page show me a list of application names".

PR #18, opened 2023-01-23
Hipp 203

tigdik:


app/controllers/CreateApplicationSuccessController.scala
```diff
@@ -0,0 +1,45 @@
...
+                                       identify: IdentifierAction,
+                                       service:ApiHubService,
+                                       val controllerComponents: MessagesControllerComponents,
+                                       view: views.html.CreateApplicationSuccessView
+                                     )(implicit ec:ExecutionContext) extends FrontendBaseController with I18nSupport {
+
+  def onPageLoad(id:String): Action[AnyContent] = identify.async {
+    implicit request => service.getApplication(id).map {
+      case Some(app) => {
+        println(app.toString)
```
PaulCDurham, line 39: Please remove the println

app/views/templates/Layout.scala.html
```diff
@@ -70,9 +71,10 @@
         ))
     }
 
-    @if(showBackLink) {
-        @govukBackLink(BackLinkViewModel(href = "#").withCssClass("js-visible"))
-    }
+
+    @if(showBackLink) {<a href="javascript:history.back()" class="govuk-back-link custom-back-link">Back</a>}
+
```
PaulCDurham, line 76: Is there a meaningful change here? If not then maybe revert to the original version.

app/views/templates/UpperPanel.scala.html
```diff
@@ -0,0 +1,21 @@
+@*
```
PaulCDurham, line 1: Is this still used?

conf/messages.en
```diff
@@ -64,3 +64,10 @@ applicationName.checkYourAnswersLabel = Application name
 applicationName.error.required = Enter an application name
 applicationName.error.length = Application name must be 100 characters or less
 applicationName.change.hidden = ApplicationName
+
+createApplicationSuccess.title = Create Application Success
+createApplicationSuccess.heading = Application created successfully
+createApplicationSuccess.h2 = What you can do now
+createApplicationSuccess.text.line1 = Your application has now been registered you can view and manage your {0}.
+createApplicationSuccess.text.line2 = Your Application ID will be available on your applications page.
+link = <a href=@routes.IndexController.onPageLoad.url class="govuk-link govuk-link--no-visited-state">@app.name
```
PaulCDurham, line 73: I don't think we should define links in the messages file, just text. If we were to add a Welsh language translation then we'd simply take all the text and translate it.

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #23, opened 2023-01-27
HIPP-289: Stride stub in local environment - Port error

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-289

This should make the login continue URL use whatever is the current port. Won't really know until this is merged and tested using Service Manager.


llmikeyj (approved):
LGTM

PR #35, opened 2023-02-17
HIPP-244: Scopes granted success page

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-244

### DIFF
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -125,3 +125,6 @@ addScope.error.length = Scope name must be 100 characters or less
 
 pendingApprovals.title = Pending production approvals
 pendingApprovals.heading = Pending production approvals
+
+scopeApproved.title = scopeApproved
+scopeApproved.heading = scopeApproved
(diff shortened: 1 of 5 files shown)


conf/messages.en
```diff
@@ -125,3 +125,8 @@ addScope.error.length = Scope name must be 100 characters or less
 
 pendingApprovals.title = Pending production approvals
 pendingApprovals.heading = Pending production approvals
+
+scopeApproved.title = Scope approved successfully
+scopeApproved.heading = Scope approved successfully
+scopeApproved.guidance.heading = What you can do now
+scopeApproved.guidance.content = Return to {0}.
```
laurahk, line 132: is this syntax `{0}` to do with navigation? Is a play thing?

laurahk (approved):

laurahk:
Only thing I noticed was an extra space on the link
link-has-space

laurahk:
Looks good to me. Thanks for the comment on the ticket about the other tickets need to be able to test the ACs.

conf/messages.en
PaulCDurham: It's a placeholder. You can call messages("page.text", placeholder) to use string substitution. In this case we want a link there so we want to substitute in the <a> element.

PR #39, opened 2023-02-24
Hipp-366

laurahk:
See comments on the ticket for background context.

Hit the frontend on the following endpoints:

- /api-hub/test-only/bad-request-demo/success
- /api-hub/test-only/bad-request-demo/user-error
- /api-hub/test-only/bad-request-demo/service-error

To test this, I have been running up api-hub-applications on port 9001 and pointing api-hub-frontend at it.


app/connectors/ApplicationsConnector.scala
```diff
@@ -17,80 +17,101 @@
...
+import play.api.libs.json.{Format, Json}
+import uk.gov.hmrc.http.{HeaderCarrier, HttpClient, HttpErrorFunctions, HttpResponse}
 import uk.gov.hmrc.play.bootstrap.config.ServicesConfig
 
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
 class ApplicationsConnector @Inject()(
-    httpClient: HttpClientV2,
+    httpClient: HttpClient,
```
laurahk, line 35: As I said on Jira, maybe we should revert the Production changes and just leave the demo and code in place ready to be used. I'm open to suggestion.

app/connectors/ConnectorErrorResponseHandling.scala
```diff
@@ -0,0 +1,97 @@
...
+      Future.failed(connectorException("No response body returned"))
+    }
+    else {
+      response.json.validate[T].fold(
+        errors => Future.failed(connectorException(s"Unable to deserialise response body: ${JsError.toJson(errors)}")),
+        t => Future.successful(t)
+      )
+    }
+  }
+
```
laurahk, line 40: I copied the function above and changed the signature to an `Either` because it's not async code and the Futures were only present for error control.

app/connectors/ConnectorErrorResponseHandling.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  def connectorException(message: String): ConnectorException = {
+    val exception = ConnectorException(message)
+    logger.error(message)
+    exception
+  }
+
+  def connectorException(response: HttpResponse): ConnectorException = {
```
laurahk, line 91: All the usages of this exception are around parsing, so I think naming it as such might be nice? Maybe just 'ResponseParseError'?

app/models/errors/ErrorResponse.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import models.Enumerable
+import play.api.libs.json.{Format, Json}
+
+sealed trait RequestError
+
+trait UserRequestError extends RequestError
+case object TeamMemberAlreadyExists extends UserRequestError
+
+trait ServiceRequestError extends RequestError
```
laurahk, line 27: As an alternative to `isRecoverable` I made 2 different types of RequestErrors so we can choose how to response to them.

app/services/BadRequestDemoService.scala
```diff
@@ -0,0 +1,36 @@
...
+
+import scala.concurrent.Future
+
+@Singleton
+class BadRequestDemoService @Inject()(badRequestDemoConnector: BadRequestDemoConnector)
+  extends Logging {
+
+  def getHardCodedApplication(respondWith: String)(implicit hc: HeaderCarrier): Future[Either[RequestError, Application]] = {
+    badRequestDemoConnector.getHardCodedApplication(respondWith)
+  }
```
laurahk, line 34: When I was duplicating the structure, I struggled to see the advantage of the Service layer in this microservice since it just calls through to the connector. Please can you explain a bit more so I understand it's use.

PR #41, opened 2023-03-02
HIPP-360: Test Stride login in Prod

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-360


app/controllers/ExperimentalStrideLoginController.scala
```diff
@@ -0,0 +1,151 @@
...
+    implicit request =>
+      logger.info("Authorising")
+      authorised(AuthProviders(PrivilegedApplication))
+        .retrieve(Retrievals.allEnrolments and Retrievals.authorisedEnrolments and Retrievals.name and Retrievals.email and Retrievals.credentials) {
+          case allEnrolments ~ authorisedEnrolments ~ name ~ email ~ credentials =>
+            Future.successful(
+              Ok(
+                view(
+                  ExperimentalStrideLoginViewModel(
+                    description = "Test login with API Hub User role required",
```
llmikeyj, line 53: We're not authorising as an api hub user in this function though, are we?

llmikeyj (commented):
Minor query about the page description

app/controllers/ExperimentalStrideLoginController.scala
PaulCDurham: Good point, I'll fix that

llmikeyj (approved):
LGTM

PR #47, opened 2023-03-10
HIPP-405: Send team members when a user clicks "Register application"

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-405


llmikeyj (approved):
LGTM.

PR #49, opened 2023-03-13
Hipp 368v

tigdik:


app/views/ConfirmAddTeamMemberView.scala.html
```diff
@@ -18,24 +18,27 @@
...
     govukRadios: GovukRadios,
     govukButton: GovukButton
 )
 
-@(form: Form[_], mode: Mode)(implicit request: Request[_], messages: Messages)
-
-@layout(pageTitle = title(form, messages("confirmAddTeamMember.title"))) {
+@(form: Form[_], teamMemberDetails: SummaryList, mode:Mode)(implicit request: Request[_], messages: Messages)
 
+@layout(pageTitle = titleNoForm(messages("confirmAddTeamMember.title")), fullWidth = true) {
```
PaulCDurham, line 28: Need to pass the user in here so that the signed in user is displayed on the page.

PaulCDurham (changes requested):

app/views/ConfirmAddTeamMemberView.scala.html
tigdik: added

PaulCDurham (approved):

PR #61, opened 2023-03-24
Hipp 462 authorise team member access

llmikeyj:


app/controllers/ApplicationDetailsController.scala
```diff
@@ -32,15 +32,11 @@ class ApplicationDetailsController @Inject()(
   identify: IdentifierAction,
```
laurahk, line 32: `ApplicationDetailsController.scala:24:17: Unused import
[warn] import services.ApiHubService`

PR #63, opened 2023-04-04
HIPP-515: primary and secondary tabs

laurahk:
I found it was not possible to have variable strings in the messages file.

I considered using the config files for QA and Prod (because those names don't change regardless of language) and using the messages file for other environments (because "primary" and "secondary" can be translated). 

However, I felt that the complexity this would bring isn't worth it considering:
* our users will only ever visit the hub on Prod
* we don't intend to swap languages using the languages file


llmikeyj (dismissed):
LGTM. Whitespace a bit tatty in EnvironmentNames

PR #64, opened 2023-04-05
hipp-400: as per prototype v5

tigdik:


PaulCDurham (changes requested):
I think we need to change the ApplicationsConnector to fetch only the applications in which the current user is a team member. At the moment it fetches all applications. I believe HIPP-399 created the endpoint to call.

app/controllers/IndexController.scala
```diff
@@ -59,4 +59,19 @@ class IndexController @Inject()(
     )
   }
 
+    def createApplication: Action[AnyContent] = identify.async { implicit request =>
```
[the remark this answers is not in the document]
laurahk: As discussed earlier, I don't believe we want both a POST and a GET for this behaviour so I think remove onSubmit method as Paul suggested.

conf/app.routes
```diff
@@ -3,6 +3,7 @@
 ->          /hmrc-frontend                                      hmrcfrontend.Routes
 
 GET         /                                                   controllers.IndexController.onPageLoad
+GET         /application-create                                 controllers.IndexController.createApplication
 POST        /                                                   controllers.IndexController.onSubmit
```
laurahk, line 7: So you'll need to remove this line too

app/controllers/IndexController.scala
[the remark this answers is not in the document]
tigdik: fixed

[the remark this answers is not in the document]
tigdik: Factored out to private method

app/connectors/ApplicationsConnector.scala
```diff
@@ -67,6 +69,18 @@ class ApplicationsConnector @Inject()(
         case Left(e) => Future.failed(e)
       }
   }
+  def getUserApplications(userEmail:String)(implicit hc: HeaderCarrier): Future[Seq[Application]] = {
```
PaulCDurham, line 72: Needs a test in ApplicationsConnectorSpec

app/services/ApiHubService.scala
```diff
@@ -55,4 +53,12 @@ class ApiHubService @Inject()(applicationsConnector: ApplicationsConnector)
     applicationsConnector.approveProductionScope(appId, scopeName)
   }
 
+  def getUserApplications(email:String)(implicit hc:HeaderCarrier): Future[Seq[Application]] = {
+    logger.debug(s"Retrieving applications for user '$email'")
```
PaulCDurham, line 57: We can't log emails for the same reason we can't use them in URLs.

PaulCDurham (changes requested):
We need to add the email address encryption key to config. That will also need to be added in higher environments too.

conf/app.routes
laurahk: I said this because I thought we agreed last week that we were getting rid of the POST endpoint all together and replacing it with a GET. So the button and the link would both be a GET.

app/connectors/ApplicationsConnector.scala
```diff
@@ -67,6 +69,18 @@ class ApplicationsConnector @Inject()(
...
+    val emailEncrypted = crypto.QueryParameterCrypto.encrypt(PlainText(userEmail)).value
```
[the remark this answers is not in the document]
tigdik: it is done automatically by RequestBuilder next line down

test/controllers/IndexControllerSpec.scala
```diff
@@ -40,14 +40,15 @@ class IndexControllerSpec extends SpecBase with MockitoSugar {
 
     "must return OK and the correct view for a GET" in {
       val applications = Seq(
-        Application("id-1", "app-name-1", Creator("creator-email-1"), Seq(TeamMember("creator-email-1"))),
-        Application("id-2", "app-name-2", Creator("creator-email-2"), Seq(TeamMember("creator-email-2")))
+        Application("id-1", "app-name-1", Creator("creator-email-2"), Seq.empty).copy(teamMembers = Seq(TeamMember("test-email"))),
+        Application("id-2", "app-name-2", Creator("creator-email-2"), Seq.empty).copy(teamMembers = Seq(TeamMember("test-email")))
```
laurahk, line 44: You don't need to copy the application again, you can use the last parameter to set the team members:
```Application("id-1", "app-name-1", Creator("creator-email-2"), Seq(TeamMember("test-email")))```
Or is there a reason you've copied it?

Also extracted "test-email" out as a `val` would help to see where it's used and why.

File: app/controllers/IndexController.scala [hunk 7]
```diff
@@ -40,13 +41,20 @@ class IndexController @Inject()(
...
+      logger.warn("Current user has no email address")
+      Future.successful(InternalServerError)
+    }(
+      email => apiHubService.getUserApplications(email).map(userApps =>
+      Ok(view(userApps, Some(request.user)))
+    ))
   }
 
-  def onSubmit: Action[AnyContent] = identify.async { implicit request =>
+  def onSubmit: Action[AnyContent] = identify.async { implicit request => createUserApplication }
```
laurahk, line 53: As I say in another comment below, I thought we had decided to have a GET, not a GET and a POST. However, the code is abstracted so I think it's fine to leave this as is 👍

test/controllers/IndexControllerSpec.scala
tigdik: Application creates teamMembers as empty collection no matter what you supply to its constructor. Check the Application def apply(...)

app/controllers/IndexController.scala
tigdik: This is what Paul suggested in his comments. With regards to having only GET - we need then to convert "create Application" button to a link which is not a hmrc component

tigdik:
> We need to add the email address encryption key to config. That will also need checking in higher environments too but I believe might be there. Check all the app-config-xxx repos for queryParameter.encryption.key.

fixed

PR #65, opened 2023-04-05
HIPP-482: primary and secondary scopes

llmikeyj:


test/controllers/AddScopeControllerSpec.scala
```diff
@@ -132,13 +138,14 @@ object AddScopeControllerSpec extends SpecBase with MockitoSugar {
...
 
-  def buildFixture(): Fixture = {
+  def buildFixture(environment1: String = "primary", environment2: String = "secondary"): Fixture = {
     val apiHubService = mock[ApiHubService]
 
     val application = applicationBuilder(userAnswers = None)
       .overrides(
         bind[ApiHubService].toInstance(apiHubService)
       )
+      .configure("environment-names.primary" -> environment1, "environment-names.secondary" -> environment2)
```
laurahk, line 148: ah nice, I didn't know you could do this! I added some code to pass in a whole Configuration object to the applicationBuilder (see https://github.com/hmrc/api-hub-frontend/blob/e36f5470a8e045c94f62f257043939193a181a43/test/base/SpecBase.scala#L48). 

I prefer your approach. Should we get rid of the code I wrote, and replace it with what you've done here?

app/models/application/EnvironmentName.scala
```diff
@@ -22,14 +22,10 @@ sealed trait EnvironmentName
...
 case object Secondary extends WithName("secondary") with EnvironmentName
-case object Dev extends WithName("dev") with EnvironmentName
-case object Test extends WithName("test") with EnvironmentName
-case object PreProd extends WithName("preProd") with EnvironmentName
-case object Prod extends WithName("prod") with EnvironmentName
 
 object EnvironmentName extends Enumerable.Implicits {
 
-  val values: Seq[EnvironmentName] = Seq(Primary, Secondary, Dev, Test, PreProd, Prod)
+  val values: Seq[EnvironmentName] = Seq(Primary, Secondary)
```
laurahk, line 28: I think scope approval will break if we remove these environments now? Is that true? (I was thinking we would play some tidy up tickets at the end of the epic).

laurahk (approved):

test/controllers/AddScopeControllerSpec.scala
laurahk: I'll do a NO-TICKET at some point to make it more consistent and get rid of my code! 😅

PaulCDurham (approved):

PR #70, opened 2023-04-18
HIPP-508: Display client ID and secret on frontend

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-508


llmikeyj (approved):
LGTM.

PR #71, opened 2023-04-24
HIPP-487: Frontend success page

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-487

### DIFF
--- a/app/controllers/GeneratePrimarySecretSuccessController.scala
+++ b/app/controllers/GeneratePrimarySecretSuccessController.scala
@@ -19,10 +19,8 @@ package controllers
 import config.FrontendAppConfig
 import controllers.actions._
 import models.application.ApplicationLenses.ApplicationLensOps
-import models.application.{Application, Credential, Secret}
+import models.application.{Application, Credential}
 import play.api.Logging
-
-import javax.inject.Inject
 import play.api.i18n.{I18nSupport, MessagesApi}
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
 import services.ApiHubService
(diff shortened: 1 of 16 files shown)


app/controllers/GeneratePrimarySecretSuccessController.scala
```diff
@@ -0,0 +1,79 @@
...
+              logger.warn(s"No primary secret generated for application $id")
+              NotFound
+          }
+        case None =>
+          logger.warn(s"Cannot find valid primary credential to generate secret for application $id")
+          Future.successful(BadRequest)
+      }
+  }
+
+  private def qaTechDeliveryValidPrimaryCredential(application: Application): Option[Credential] = {
```
laurahk, line 68: Why is this prefixed with 'qaTechDelivery`?

test/controllers/GeneratePrimarySecretSuccessControllerSpec.scala
```diff
@@ -0,0 +1,148 @@
...
+      val fixture = buildFixture()
+
+      running(fixture.playApplication) {
+        when(fixture.apiHubService.getApplication(ArgumentMatchers.eq(FakeApplication.id))(any()))
+          .thenReturn(Future.successful(None))
+
+        val request = FakeRequest(GET, routes.GeneratePrimarySecretSuccessController.onPageLoad(FakeApplication.id).url)
+        val result = route(fixture.playApplication, request).value
+
+        status(result) mustBe NOT_FOUND
```
laurahk, line 84: Doesn't need to be done as part of this ticket, but I'm thinking it would be better to redirect the user to a more user-friendly error message page rather than the browser rendered 400. What do you think?

PaulCDurham: I think we should eventually do that in multiple error scenarios. Something we could parameterise with context would be good.

laurahk (approved):
A chunky PR! Looks fab 😃

PR #73, opened 2023-05-04
HIPP-559: Remove old environments from data model

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-559


llmikeyj (approved):
LGTM! Happy to see those go..

PR #76, opened 2023-05-16
hipp-531: removed optionality of field 'email' in sign in form

tigdik:


app/testonly/Retrievals.scala
```diff
@@ -34,7 +34,7 @@ object Retrievals {
       "retrievals" -> JsArray(
         Seq(
           JsString(retrievals.principal),
-          retrievals.email.map(JsString).getOrElse(JsNull),
+          JsString(retrievals.principal),
```
PaulCDurham, line 37: I think this needs to be email not principal.

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #78, opened 2023-05-22
Hipp 600

tigdik:


test/controllers/IndexControllerSpec.scala
```diff
@@ -135,4 +184,16 @@ object IndexControllerSpec extends SpecBase with MockitoSugar {
     Fixture(application, mockSessionRepository, mockApiHubService)
   }
 
+  def buildFixtureWithUser(userModel: UserModel = FakeUser): Fixture = {
```
PaulCDurham, line 187: This basically repeats buildFixture. We have several examples of buildFixture that have a "userModel: UserModel = FakeUser" argument. I'd simply copy that pattern. As the argument has a default value it will work with the existing tests.

PaulCDurham (approved):

test/controllers/IndexControllerSpec.scala
tigdik: fixed

PR #79, opened 2023-05-30
HIPP-605: Create frontend “Delete Application” button

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-605

### DIFF
--- a/it/connectors/ApplicationsConnectorSpec.scala
+++ b/it/connectors/ApplicationsConnectorSpec.scala
@@ -136,6 +136,42 @@ class ApplicationsConnectorSpec
     }
   }
 
+  "ApplicationsConnector.deleteApplication" - {
+    "must place the correct request" in {
+      val id = "test-id"
+
+      stubFor(
+        delete(urlEqualTo(s"/api-hub-applications/applications/$id"))
+          .withHeader("Authorization", equalTo("An authentication token"))
+          .willReturn(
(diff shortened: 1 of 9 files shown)


test/services/ApiHubServiceSpec.scala
```diff
@@ -104,6 +104,40 @@ class ApiHubServiceSpec extends AsyncFreeSpec with Matchers with MockitoSugar wi
...
+
+      service.deleteApplication(id)(HeaderCarrier()) map {
+        actual =>
+          actual mustBe Some(())
+          verify(applicationsConnector).deleteApplication(any())(any())
+          succeed
+      }
+    }
+
+    "must return None when the applications connectors does to indicate the application was not found" in {
```
laurahk, line 125: think "does to" needs removing, but perhaps get that on our next visit to the file

it/connectors/ApplicationsConnectorSpec.scala
```diff
@@ -136,6 +136,42 @@ class ApplicationsConnectorSpec
...
+          )
+      )
+
+      buildConnector(this).deleteApplication(id)(HeaderCarrier()) map {
+        actual =>
+          actual mustBe None
+      }
+    }
+  }
+
```
laurahk, line 174: How come in this file (across all the tests) we don't test the Left -> Failed scenario?

PR #84, opened 2023-06-27
HIPP-588: Test connectivity page

llmikeyj:


app/connectors/ApplicationsConnector.scala
```diff
@@ -139,4 +140,17 @@ class ApplicationsConnector @Inject()(
       }
   }
 
+  def testConnectivity()(implicit hc:HeaderCarrier): Future[String] = {
+    httpClient
+      .get(url"$applicationsBaseUrl/api-hub-applications/test-connectivity")
```
PaulCDurham, line 145: Needs an Authorization header as the backend endpoint is authenticated

PaulCDurham (changes requested):
Just needs that auth header when calling the backend

app/connectors/ApplicationsConnector.scala
llmikeyj: Ok

PaulCDurham (approved):

PR #90, opened 2023-07-11
Hipp 606 google analytics

llmikeyj:


app/views/templates/Layout.scala.html
```diff
@@ -89,7 +90,7 @@
 
 @govukLayout(
     pageTitle   = Some(pageTitle),
-    headBlock   = Some(head),
+    headBlock   = Some(hmrcInternalHead()),
```
laurahk, line 93: I think this should be ```Some(hmrcInternalHead(head))``` (see [HmrcInternalHead](https://github.com/hmrc/play-frontend-hmrc/blob/a68ac35a51d164c104a66085a138203611dab2c7/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/helpers/HmrcInternalHead.scala.html#L23)) otherwise we are changing functionality by not including the `head` bit anymore (see [Layout.scala.html#L42](https://github.com/hmrc/api-hub-frontend/blob/main/app/views/templates/Layout.scala.html#L42)) and not dealing with timeouts. We don't actually deal with timeouts properly yet but I believe this is coming.

app/views/templates/ApiHubHeader.scala.html
```diff
@@ -21,6 +21,11 @@
 )
 
 @(showBackLink: Boolean, user: Option[UserModel])(implicit request: Request[_], messages: Messages)
+<!-- Google Tag Manager (noscript) -->
```
laurahk, line 24: So step 3 on the ticket says: 

> Add the noscript snippet to ApiHubHeader (as we don't use the standard HmrcInternalHeader which would normally do this for us). Details are in the above guidelines.

But I can't see why we need to do this. I can't find any documentation that says we should. [HmrcInternalHeader](https://github.com/hmrc/play-frontend-hmrc/blob/a68ac35a51d164c104a66085a138203611dab2c7/src/main/twirl/uk/gov/hmrc/hmrcfrontend/views/components/HmrcInternalHeader.scala.html) doesn't look like it does anything with GTM

laurahk: Given it says it should be the first tag after the body, it feels weird that the noscript tag would live in a file called "...Header". Saying that I don't understand enough about Twirl and the GovUkLayout to know where else it should go.

app/views/templates/Layout.scala.html
laurahk: Yeah without the head, if you run up the app it looks like this:

laurahk: This brings back the tracking consent error in the Console - but I think we can just get rid of 
```
tracking-consent-frontend {
  gtm.container = "transitional"
}
```
in application.conf. Not entirely sure what the implications of that are though.

laurahk: Only info I can find is in the [decision log for tracking-consent-frontend](https://github.com/hmrc/tracking-consent-frontend/blob/fa733daec936c772832ac7e2b1dc101d8450e3e1/docs/adr/0004-namespace-the-gtm-container-attribute-for-tracking-consent.md?plain=1#L10). I mean we aren't doing tracking consent so I think we can safely get rid of it.

laurahk (changes requested):
* Ensure the app looks correct by running it up locally - see my comment about including head (I _think_ this is the solution)
* Ensure the tracking consent error is not displayed
* Try and get the noscript tag to be first in the body - perhaps it doesn't matter, not sure.

app/views/templates/Layout.scala.html
llmikeyj: Okay will do, thanks

app/views/templates/ApiHubHeader.scala.html
llmikeyj: Sorry, me neither. Shall we leave it in for now?

laurahk: It's not actually the first tag after the body tag. Do you think that matters? I guess (once deployed) we could test it by turning JS off in the browser and then doing something and see if the analytics sends.

laurahk: I think leave it where it is for now.

laurahk (approved):

PR #98, opened 2023-09-05
Hipp 719 remove dead footer links

llmikeyj:


app/views/templates/Layout.scala.html
```diff
@@ -66,6 +66,11 @@
         customScriptsBlock.get
         }
     }
+    <script @{CSPNonce.attr} defer="true">
+    for (var i = 0; i < 3; i++) {
```
edgarjimenez, line 70: I can't see how we can be sure the right thing is removed with this approach. It would be nice to add a test to confirm this works.

I think we should be able to initialise the `govukLayout` with a custom `footerBlock` as done [here](https://github.com/hmrc/cip-fph-developer-frontend/blob/main/app/uk/gov/hmrc/cip/fph/views/global/components/Layout.scala.html#L121). This approach would require a bit more work, as the footer's URLs would have to be configured per environment. However, we would be in total control of what is included in the footer.
Happy to have a conversation irl

llmikeyj: Ok

edgarjimenez (approved):

PR #100, opened 2023-09-12
HIPP-727: ATI 4.0 Investigate parsing errors

PaulCDurham:


app/config/FrontendAppConfig.scala
```diff
@@ -32,7 +32,7 @@ class FrontendAppConfig @Inject() (configuration: Configuration) {
   private val contactFormServiceIdentifier = "api-hub-frontend"
 
   def feedbackUrl(implicit request: RequestHeader): String =
-    s"$contactHost/contact/beta-feedback?service=$contactFormServiceIdentifier&backUrl=${SafeRedirectUrl(host + request.uri).encodedUrl}"
+    s"$contactHost/contact/beta-feedback?service=$contactFormServiceIdentifier&backUrl=${url"$host${request.uri}"}"
```
PaulCDurham, line 35: Dependency upgrade deprecated SafeRedirectUrl and this is the recommended replacement

app/views/util/FooterWithItems.scala.html
```diff
@@ -39,4 +39,4 @@
 }
 
 
-@footer(new Footer(meta = Some(Meta(items = Some(footerlinkItems)))))
+@footer(new Footer(meta = Some(Meta(items = Some(footerlinkItems())))))
```
PaulCDurham, line 42: Fixes compiler warning

build.sbt
```diff
@@ -68,7 +68,8 @@ lazy val root = (project in file("."))
     Assets / pipelineStages := Seq(concat,uglify),
     // only compress files generated by concat
     uglify / includeFilter := GlobFilter("application.js"),
-    Global / excludeLintKeys += update / evictionWarningOptions
+    Global / excludeLintKeys += update / evictionWarningOptions,
+    scalacOptions ++= Seq("-deprecation", "-feature")
```
PaulCDurham, line 72: We now get details of deprecation warnings rather than simply told we have them

test-utils/generators/ModelGenerators.scala
```diff
@@ -53,4 +53,13 @@ trait ModelGenerators {
...
 
+  lazy val genLegalUnicodeChar: Gen[Char] = Arbitrary.arbChar.arbitrary.suchThat(
+    c =>
+      Character.isDefined(c)
+        && Character.UnicodeBlock.of(c) != Character.UnicodeBlock.PRIVATE_USE_AREA
+        && !Character.isISOControl(c)
+  )
+
+  lazy val genLegalUnicodeString: Gen[String] = Gen.nonEmptyListOf(genLegalUnicodeChar).map(_.mkString)
+
```
PaulCDurham, line 64: We used to have arbitrary string but that fails HTML validation as it uses characters that cannot appear in HTML documents. This restricts the characters used so that they are legal.

edgarjimenez (approved):

PR #120, opened 2023-10-13
HIPP-782: Manage application APIs - No APIs

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-782

### DIFF
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -117,6 +117,12 @@ applicationDetails.generateSecret = Generate secret
 applicationDetails.deleteApplication = Delete application
 applicationDetails.thereWereIssues = There were issues fetching elements of this application
 
+applicationApis.title = Application APIs
+applicationApis.heading = Manage application API''s
+applicationApis.apis.heading = Application API''s
+applicationApis.apis.hipApis = add API''s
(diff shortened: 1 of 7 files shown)


conf/messages.en
```diff
@@ -117,6 +117,12 @@ applicationDetails.generateSecret = Generate secret
 applicationDetails.deleteApplication = Delete application
 applicationDetails.thereWereIssues = There were issues fetching elements of this application
 
+applicationApis.title = Application APIs
+applicationApis.heading = Manage application API''s
+applicationApis.apis.heading = Application API''s
+applicationApis.apis.hipApis = add API''s
+applicationApis.apis.noApis = You have no API''s added, {0} to your application.
```
edgarjimenez, line 124: This looks like it should be be split into two separate messages

PaulCDurham: We could do it either way but this is how we've previously implemented a link within a sentence. In this case applicationApis.apis.noApis is the sentence and applicationApis.apis.hipApis is the hyperlink text. If we split the messages we'd then have three entries: sentence part 1, link text, sentence part 2.

edgarjimenez: I completely misunderstood the structure of that message 👍

edgarjimenez (approved):

PR #121, opened 2023-10-13
HIPP-774: Display apis alphabetically

llmikeyj:


app/controllers/HipApisController.scala
```diff
@@ -38,7 +38,7 @@ class HipApisController @Inject()(
   def onPageLoad(): Action[AnyContent] = optionallyIdentified.async {
     implicit request =>
       apiHubService.getAllHipApis().map {
-        case apiDetails: Seq[ApiDetail] => Ok(view(request.user, apiDetails))
+        case apiDetails: Seq[ApiDetail] => Ok(view(request.user, apiDetails.sortWith( _.title < _.title)))
```
PaulCDurham, line 41: In the technical details we specify a case-insensitive sort but this isn't reflected in the ACs which simply state alphabetic. Maybe double-check that with Whitney? If it's case-insensitive then maybe add a differentiating example to the test below, eg upperCaseZebra.

PaulCDurham (commented):
Please check on case-insensitive question with Whitney

app/controllers/HipApisController.scala
llmikeyj: Okay

PaulCDurham (approved):

PR #126, opened 2023-10-20
HIPP-882: Application apis

llmikeyj:


PaulCDurham (changes requested):
You're missing this bit from the Jira ticket. We're calling addScopes at the end of the Add an API journey. We want to call the new addApi endpoint. You'll need to add that in to the connector and service layers and then call it from AddAnApiCompleteController.addApi().

"Change the FE to use the new endpoint described above to add an API, instead of just Add Scope as it currently does"

test/controllers/AddAnApiCompleteControllerSpec.scala
```diff
@@ -49,12 +50,24 @@ class AddAnApiCompleteControllerSpec extends SpecBase with HtmlValidation {
     .set(AddAnApiSelectEndpointsPage, selectedScopes).toOption.value
     .set(ApiPolicyConditionsDeclarationPage, acceptedPolicyConditions).toOption.value
 
+  private val apiDetail = mock[ApiDetail]
```
PaulCDurham, line 53: I don't get mocking a case class. Just set one up using the constructor.

app/models/application/Application.scala
```diff
@@ -28,7 +28,8 @@ case class Application (
   lastUpdated: LocalDateTime,
   teamMembers: Seq[TeamMember],
   environments: Environments,
-  issues: Seq[String] = Seq.empty
+  issues: Seq[String] = Seq.empty,
+  apis: Seq[Api] = Seq.empty
```
PaulCDurham, line 32: Don't think we need the default empty. We really want one from the backend.

app/services/ApiHubService.scala
```diff
@@ -57,6 +60,12 @@ class ApiHubService @Inject()(
     applicationsConnector.addScopes(id, scopeNames.map(name => NewScope(name, Seq(Secondary))).toSeq)
   }
 
+  def addApi(applicationId: String, apiId: String, availableEndpoints: Seq[AvailableEndpoint])(implicit hc: HeaderCarrier): Future[Option[Unit]] = {
+    val scopes = availableEndpoints.flatMap(ae => ae.endpointMethod.scopes)
+    val endpoints = availableEndpoints.map(ae => AddApiRequestEndpoint(ae.endpointMethod.httpMethod, ae.path))
+    applicationsConnector.addApi(applicationId, AddApiRequest(apiId, endpoints, scopes))
+  }
```
PaulCDurham, line 67: Could you add a unit test to ApiHubServiceSpec? Need to make sure we're grabbing the correct endpoints and scopes.

PR #132, opened 2023-11-02
HIPP-838: Add new credential

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-838


app/views/application/EnvironmentAndCredentialsView.scala.html
```diff
@@ -87,9 +106,49 @@ <h2 class="govuk-heading-m">@messages("environmentAndCredentials.development.hea
...
+                    ),
+                    TableRow(
+                        content = Text(credential.secretFragment.map(secret => s"****$secret").getOrElse(""))
+                    ),
+                    TableRow(
+                        content = Text(ViewUtils.formatShortLocalDateTime(credential.created))
+                    )
+                )
+        ),
+        head = Some(Seq(
```
edgarjimenez, line 139: should we hide this headers if there are no credentials to display?

PaulCDurham: I discussed that with Jonathon during UI review as there are no ACs on this, only hiding the first production credential. He's going to chat to Whitney and maybe add a ticket added to do something like that.

it/connectors/ApplicationsConnectorSpec.scala
```diff
@@ -363,6 +368,63 @@ class ApplicationsConnectorSpec
...
+          )
+      )
+
+      buildConnector(this).addCredential(FakeApplication.id, Primary)(HeaderCarrier()) map {
+        actual =>
+          actual mustBe Right(None)
+      }
+    }
+
+    "must Conflict???" in {
```
edgarjimenez, line 410: Are we not sure it should be 409?

edgarjimenez (approved):

PR #141, opened 2023-11-23
Hipp 870 request production access success

llmikeyj:


app/controllers/application/RequestProductionAccessEndJourneyController.scala
```diff
@@ -0,0 +1,136 @@
...
+    userAnswers.get(ProvideSupportingInformationPage) match {
+      case Some(information) => Right(information)
+      case None => Left(controllers.application.routes.ProvideSupportingInformationController.onPageLoad())
+    }
+  }
+
+  private def buildAccessRequest(application: Application, supportingInformation: String, requestedBy: String)(implicit request: DataRequest[AnyContent]): Future[Either[Result, AccessRequestRequest]] = {
+    applicationApiBuilder.build(application).map {
+      case Right(applicationApis) =>
+        val accessRequestApis = applicationApis.map(applicationApi => {
```
PaulCDurham, line 121: I think we need some filtering here. If we go through multiple loops of add API/add endpoints etc then we don't always want to add all endpoints of all APIs.

applicationApis.filter(_.endpoints.exists(_.primaryAccess == Inaccessible)).map(... etc etc

app/controllers/application/RequestProductionAccessEndJourneyController.scala
```diff
@@ -0,0 +1,136 @@
...
+          val accessRequestEndpoints = applicationApi.endpoints.map(endpoint => AccessRequestEndpoint(endpoint.httpMethod, endpoint.path, endpoint.scopes))
```
PaulCDurham, line 122: Similarly we only want inaccessible endpoints.

applicationApi.endpoints.filter(_.primaryAccess == Inaccessible).map(... etc

app/controllers/application/RequestProductionAccessEndJourneyController.scala
llmikeyj: These apis are generated using the same code as for the request production access page; see 

RequestProductionAccessController
```

Is this also wrong?
```

app/views/application/RequestProductionAccessSuccessView.scala.html
```diff
@@ -0,0 +1,90 @@
...
+                value = Value(
+                    content = HtmlContent(app.name)
+                ),
+            ),
+            SummaryListRow(
+                key = Key(
+                    content = Text(messages("requestProductionAccessSuccess.apiName"))
+                ),
+                value = Value(
+                    content = Text(apis.head.apiName)
```
PaulCDurham: Could you have a chat with Jonathon about adding multiple APIs? Perhaps in that scenario we should display all the API names. Perhaps add that as a UI improvement ticket?

PaulCDurham (changes requested):
One thing to fix, one question for the review with Jonathon

app/views/application/RequestProductionAccessSuccessView.scala.html
PaulCDurham: I see you've done that, cool

PaulCDurham (approved):

PaulCDurham (approved):

PR #146, opened 2023-11-28
HIPP-877: Reject production access request

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-877

### DIFF
--- a/app/controllers/admin/AccessRequestController.scala
+++ b/app/controllers/admin/AccessRequestController.scala
@@ -92,7 +92,7 @@ class AccessRequestController @Inject()(
   private def reject(id: String, rejectedReason: String)(implicit request: IdentifierRequest[AnyContent]): Future[Result] = {
     request.user.email match {
       case Some(email) =>
-        apiHubService.rejectAccessRequest(id, email, rejectedReason).map {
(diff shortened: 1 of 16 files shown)


app/controllers/admin/AccessRequestController.scala
```diff
@@ -79,15 +82,22 @@ class AccessRequestController @Inject()(
...
   }
 
-  private def reject(id: String)(implicit request: IdentifierRequest[AnyContent]): Future[Result] = {
-    buildView(id, OK, form)
+  private def reject(id: String, rejectedReason: String)(implicit request: IdentifierRequest[AnyContent]): Future[Result] = {
+    request.user.email match {
+      case Some(email) =>
+        apiHubService.rejectAccessRequest(id, email, rejectedReason.trim).map {
+          case Some(_) => Ok(rejectedSuccessView(request.user))
+          case _ => accessRequestNotFound(id)
```
llmikeyj, line 97: Aren't we throwing away the conflict scenario I mentioned on the back end ticket here?

PaulCDurham: Yes, effectively. We'd need to work out how to ripple that up to the controller layer. We've considered that before I think and not done anything after some experiments. Perhaps something to chat about.

PR #152, opened 2023-12-13
HIPP-917: Modify Select Endpoints Page

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-917

There's some refactoring in this PR. Both the API Id and Application Id in user answers have been replaced with the objects rather than identifiers. This is because we need access to them much more often to identify which endpoints have been added previously, ie in a prior add an API or add endpoints journey.

### DIFF
--- a/app/controllers/AddAnApiSelectEndpointsController.scala
+++ b/app/controllers/AddAnApiSelectEndpointsController.scala
@@ -17,15 +17,17 @@
 package controllers
 
 import controllers.actions._
-import controllers.helpers.ErrorResultBuilder
 import forms.AddAnApiSelectEndpointsFormProvider
-import models.{AddAnApiContext, Mode}
+import models.api.ApiDetail
+import models.application.Application
+import models.requests.DataRequest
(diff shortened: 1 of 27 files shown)


app/controllers/AddAnApiSelectEndpointsController.scala
```diff
@@ -52,54 +51,54 @@ class AddAnApiSelectEndpointsController @Inject()(
...
-      )
-    )
+  private def formDataWithExistingEndpoints(request: DataRequest[AnyContent], apiDetail: ApiDetail, application: Application): Map[String, Seq[String]] = {
+
+    // We are disabling the checkboxes for endpoints that have already been
+    // added. Disabled inputs are not submitted with the other form data so we
+    // lose the data associated with them when this is posted back to us.
+    // So we need to add the already-added endpoints into the form data.
+
+    request.request.body.asFormUrlEncoded.getOrElse(Map.empty) ++
```
edgarjimenez, line 94: 👍

edgarjimenez (approved):

llmikeyj (approved):

PR #187, opened 2024-03-22
HIPP-1078 - updating fields in GenerateResponse

rdawson-hmrc:


app/models/deployment/GenerateResponse.scala
```diff
@@ -20,7 +20,7 @@ import play.api.libs.json.{Format, Json}
 
 sealed trait GenerateResponse
 
-case class SuccessfulGenerateResponse(projectId: Int, lineOfBusiness: String, branchName: String, mergeRequestIid: Int) extends GenerateResponse
+case class SuccessfulGenerateResponse(id: String, version: String, mergeRequestIid: Int, uri: String) extends GenerateResponse
```
PaulCDurham, line 23: Should we change the name of this to match the BE?

PaulCDurham: Ditto with other classes like GenerateRequest

edgarjimenez (approved):

PR #196, opened 2024-04-11
HIPP-1045 - Manage team members

rdawson-hmrc:


app/viewmodels/team/ManageTeamMembers.scala
```diff
@@ -0,0 +1,58 @@
...
+import uk.gov.hmrc.govukfrontend.views.viewmodels.content.Text
+import uk.gov.hmrc.govukfrontend.views.viewmodels.summarylist.{Actions, Key, SummaryListRow}
+import viewmodels.govuk.summarylist._
+import viewmodels.implicits._
+
+object ManageTeamMembers {
+
+  def rows(teamMembers: Seq[TeamMember])(implicit messages: Messages): Seq[SummaryListRow] = {
+    val teamMembersWithIndexes = teamMembers.zipWithIndex
+    val (currentUser, otherTeamMembers) = (teamMembersWithIndexes.head, teamMembersWithIndexes.tail)
```
edgarjimenez, line 30: how is this order guaranteed? i.e. how do we know the head is always the current user? I might be missing something obvious here

rdawson-hmrc: When we first create the UserAnswers in CreateTeamStartController we immediately populate it with the current user's email address, then any other email addresses get appended to the Seq, so as I understand it that means the current user will always be at the head of the Seq

rdawson-hmrc: If you think that assumption isn't safe I can do a 'find' on the Seq and pull out the current users email wherever it is, it's just a bit more code to do it that way

app/controllers/team/ManageTeamMembersController.scala
```diff
@@ -0,0 +1,61 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
```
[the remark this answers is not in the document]
edgarjimenez: Just realised that there are still a lot of files with 2023. I thought we had updated all of them before.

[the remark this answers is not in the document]
rdawson-hmrc: Do we need to change old files (ie files created in past years?) - this suggests it's actually a bad idea to keep bringing the date forward: https://stackoverflow.com/a/2391555/138256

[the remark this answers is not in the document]
edgarjimenez: That's a fair point. I've seen the date being moved forward in the past by multiple teams in HMRC, but it's probably worth asking the question, just to make sure what the HMRC official approach is.
Let's leave it as is, we can come back to this later.

app/viewmodels/team/ManageTeamMembers.scala
edgarjimenez: I think `ManageTeamMembers.rows` shouldn't take that ordering for granted, so adding something to identify the current user would be better, I think.
An option could be to make `rows` take the current user as a parameter. The controller should be able to extract it from the request, but you may have a better/simpler way of doing it.

app/controllers/team/ManageTeamMembersController.scala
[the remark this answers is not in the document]
edgarjimenez: It looks like we just have to make sure to add the correct copyright date for new files. Older files should stay as they are, unless they've been significantly changed.

test/controllers/team/ManageTeamMembersControllerSpec.scala
```diff
@@ -0,0 +1,105 @@
...
+      running(fixture.application) {
+        val request = FakeRequest(routes.ManageTeamMembersController.onPageLoad())
+        val result = route(fixture.application, request).value
+
+        status(result) mustBe SEE_OTHER
+        redirectLocation(result) mustBe Some(controllers.routes.JourneyRecoveryController.onPageLoad().url)
+      }
+    }
+
+    "must display error page if current user has no email" in {
```
edgarjimenez, line 65: 👍

edgarjimenez (approved):

PR #227, opened 2024-05-15
HIPP-1221 add domain/subdomain data in frontend

rdawson-hmrc:


app/models/api/Domain.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package models.api
+
+case class Domain(code: String, description: String, subDomains: Seq[SubDomain])
+case class SubDomain(code: String, description: String)
+
+class Domains(val domains: Seq[Domain]) {} // add methods for looking up descriptions from codes in here
```
PaulCDurham, line 22: See comment above. I'd make this a component that provides domain data. We can then add useful utility methods as you suggest.

app/services/ApiHubService.scala
```diff
@@ -139,4 +141,5 @@ class ApiHubService @Inject()(
     applicationsConnector.addTeamMemberToTeam(id, teamMember)
   }
 
+  def getDomains(): Domains = new Domains(config.domains)
```
PaulCDurham, line 144: See above. Inject Domains and get domain data from that.

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #242, opened 2024-05-24
HIPP 1236 api auth action

rdawson-hmrc:


app/controllers/actions/ApiAuthAction.scala
```diff
@@ -0,0 +1,79 @@
...
+  errorResultBuilder: ErrorResultBuilder,
+  override val messagesApi: MessagesApi
+) extends ApiAuthActionProvider with I18nSupport {
+
+  def apply(apiId: String)(implicit ec: ExecutionContext): ApiAuthAction = {
+    new ApiAuthAction with FrontendHeaderCarrierProvider {
+      override protected def refine[A](identifierRequest: IdentifierRequest[A]): Future[Either[Result, ApiRequest[A]]] = {
+        implicit val request: Request[_] = identifierRequest
+
+        apiHubService.getApiDetail(apiId).flatMap(_ match {
```
PaulCDurham, line 48: Fyi you can simply omit the match inside flatMap. That can be easier to read: flatMap { case this => etc case that => etc }

test/controllers/actions/FakeUser.scala
```diff
@@ -25,3 +25,5 @@ object FakeUserLdapNoEmail extends UserModel("id", "test-name", LdapUser, None)
 object FakeUserStrideNoEmail extends UserModel("id", "test-name", StrideUser, None)
 
 object FakeUserNotTeamMember extends UserModel("idx", "test-namex", LdapUser, Some("different-email"))
+
+object FakeUserWithSupportRole extends UserModel("id", "support-user", LdapUser, Some("support-user@example.com"), Permissions(false, true, false))
```
PaulCDurham, line 29: There is already a controllers.actions.FakeSupporter. Could we stick with that one?

test/controllers/actions/ApiAuthActionSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+          )(FakeRequest(), messages(fixture.playApplication)).toString()
+      }
+    }
+
+    "if API exists and user has the support role then it returns OK" in {
+      val fixture = buildFixture()
+
+      when(fixture.apiHubService.getApiDetail(ArgumentMatchers.eq(FakeApiDetail.id))(any())).thenReturn(Future.successful(Some(FakeApiDetail)))
+      when(fixture.apiHubService.findTeams(ArgumentMatchers.eq(FakeUserWithSupportRole.email))(any())).thenReturn(
+        Future.successful(Seq(Team(FakeApiDetail.teamId.get, "team name", LocalDateTime.now(), Seq.empty)))
```
PaulCDurham, line 69: We shouldn't need this stub. You could almost argue we should verify that this method is not called if the user is support.

test/controllers/actions/ApiAuthActionSpec.scala
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    "if API exists but is not owned by any team then it returns Unauthorized" in {
+      val fixture = buildFixture()
+
+      when(fixture.apiHubService.getApiDetail(ArgumentMatchers.eq(FakeApiDetail.id))(any())).thenReturn(
+        Future.successful(Some(FakeApiDetail.copy(teamId = None)))
+      )
+      when(fixture.apiHubService.findTeams(ArgumentMatchers.eq(FakeUser.email))(any())).thenReturn(
+        Future.successful(Seq(Team("teamId", "team name", LocalDateTime.now(), Seq.empty)))
```
PaulCDurham, line 86: Again, we should not need this stub and maybe verify this method isn't called instead.

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #246, opened 2024-05-30
HIPP-1274: Promote to production

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1274

### DIFF
--- a/app/views/myapis/SimpleApiPromotionView.scala.html
+++ b/app/views/myapis/SimpleApiPromotionView.scala.html
@@ -1,0 +1,72 @@
+@*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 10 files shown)


app/views/myapis/SimpleApiPromotionView.scala.html
```diff
@@ -0,0 +1,72 @@
...
+    govukSummaryList: GovukSummaryList,
+    formHelper: FormWithCSRF,
+    govukButton : GovukButton
+)
+
+@(apiDetail: ApiDetail, user: UserModel)(implicit request: Request[_], messages: Messages)
+
+@layout(pageTitle = titleNoForm("Promote to production"), user = Some(user)) {
+
+    <h1 class="govuk-heading-l">You are about to promote your API to production</h1>
```
rdawson-hmrc, line 31: should these be extracted to the messages file?

PaulCDurham: Normally yes, but in the temporary pages we've taken to simply using text. Saves tidying up all the messages later.

rdawson-hmrc (approved):

PR #253, opened 2024-06-07
HIPP-1290 - adding hods and domains to api update form

rdawson-hmrc:


PaulCDurham (changes requested):
Could you add the new mandatory fields to the tests in "must return 400 Bad Request and errors when invalid data is submitted"?

PaulCDurham (approved):

PR #278, opened 2024-07-18
HIPP-1349 change owning team

rdawson-hmrc:


app/controllers/myapis/UpdateApiTeamController.scala
```diff
@@ -0,0 +1,66 @@
...
+)(implicit ec: ExecutionContext) extends FrontendBaseController with I18nSupport {
+  private val form = formProvider()
+
+  def onPageLoad(apiId: String): Action[AnyContent] = (identify andThen apiAuth(apiId)) async {
+    implicit request => showView(OK, form)
+  }
+
+  private def showView(code: Int, form: Form[_])(implicit request: ApiRequest[_]): Future[Result] = {
+    apiHubService.findTeams(None).map(teams => {
+      val owningTeam = teams.find(team => request.apiDetails.teamId.contains(team.id))
```
llmikeyj, line 50: Why wouldn't we use apiHubService.findTeamById here?

rdawson-hmrc: We need the full list of teams to pass to the view anyway, so filtering like this just avoids a second API call

llmikeyj (approved):

PR #282, opened 2024-07-22
HIPP-1371: Rewire change api team to go via applications

llmikeyj:


app/connectors/ApplicationsConnector.scala
```diff
@@ -441,6 +441,17 @@ class ApplicationsConnector @Inject()(
...
 
+  def updateApiTeam(apiId: String, teamId: String)(implicit hc: HeaderCarrier): Future[ApiDetail] = {
+    httpClient.put(url"$applicationsBaseUrl/api-hub-applications/apis/$apiId/teams/$teamId")
+      .setHeader((ACCEPT, JSON))
+      .setHeader(AUTHORIZATION -> clientAuthToken)
+      .execute[Either[UpstreamErrorResponse, ApiDetail]]
+      .flatMap {
+        case Right(apiDetail) => Future.successful(apiDetail)
+        case Left(e) => Future.failed(e)
+      }
```
PaulCDurham, line 452: The controller can return 404 Not Found - could we add support and a test for that?

PaulCDurham (changes requested):

app/connectors/ApplicationsConnector.scala
```diff
@@ -441,13 +441,14 @@ class ApplicationsConnector @Inject()(
...
-  def updateApiTeam(apiId: String, teamId: String)(implicit hc: HeaderCarrier): Future[ApiDetail] = {
+  def updateApiTeam(apiId: String, teamId: String)(implicit hc: HeaderCarrier): Future[Either[ApplicationsException, Unit]] = {
     httpClient.put(url"$applicationsBaseUrl/api-hub-applications/apis/$apiId/teams/$teamId")
       .setHeader((ACCEPT, JSON))
       .setHeader(AUTHORIZATION -> clientAuthToken)
       .execute[Either[UpstreamErrorResponse, ApiDetail]]
       .flatMap {
-        case Right(apiDetail) => Future.successful(apiDetail)
+        case Right(_) => Future.successful(Right())
+        case Left(e) if e.statusCode == NOT_FOUND => Future.successful(Right())
```
PaulCDurham, line 451: Maybe change the return signature to Option[Unit] and return None for Not Found? At the moment this is returning success when the BE says failure. No Left value is ever returned.

PaulCDurham (changes requested):

app/connectors/ApplicationsConnector.scala
llmikeyj: Maybe. I think the issue is though that we don't have a defined use case for this scenario. All of the other connectors ignore Not Founds I think, so that the FE keeps going.

PaulCDurham: This is from getApplication:
`case Left(e) if e.statusCode==404 => Future.successful(None)`

The controller should probably show an error page with a message like "API or Team not found".

There should be ACs for this really but they're missing.

app/connectors/ApplicationsConnector.scala
```diff
@@ -441,6 +441,18 @@ class ApplicationsConnector @Inject()(
       .execute[Seq[UserContactDetails]]
   }
 
+  def updateApiTeam(apiId: String, teamId: String)(implicit hc: HeaderCarrier): Future[Either[ApplicationsException, Option[Unit]]] = {
```
PaulCDurham, line 444: Just Future[Option[Unit]]? The code doesn't return a Left value so the Either is redundant.

app/connectors/ApplicationsConnector.scala
llmikeyj: Done

app/connectors/ApplicationsConnector.scala
llmikeyj: Ok

PaulCDurham (approved):

PR #296, opened 2024-08-08
HIPP-1402: Add egressPrefix & prefixesToRemove

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1402

### DIFF
--- a/app/controllers/myapis/SimpleApiDeploymentController.scala
+++ b/app/controllers/myapis/SimpleApiDeploymentController.scala
@@ -24,7 +24,7 @@ import forms.mappings.Mappings
 import models.deployment.{DeploymentsRequest, InvalidOasResponse, SuccessfulDeploymentsResponse}
 import models.requests.IdentifierRequest
 import play.api.Logging
-import play.api.data.Forms.mapping
+import play.api.data.Forms.{mapping, optional}
 import play.api.data._
 import play.api.i18n.I18nSupport
 import play.api.libs.json.Json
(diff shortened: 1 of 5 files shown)


app/controllers/myapis/SimpleApiDeploymentController.scala
```diff
@@ -102,10 +102,24 @@ object SimpleApiDeploymentController {
...
         "domain" -> text("Enter a domain"),
         "subdomain" -> text("Enter a subdomain"),
-        "hods" -> Forms.seq(text())
+        "hods" -> Forms.seq(text()),
+        "prefixesToRemove" -> optional(text()).transform[Seq[String]](toPrefixesToRemove, fromPrefixesToRemove),
+        "egressPrefix" -> optional(text())
         )(DeploymentsRequest.apply)(DeploymentsRequest.unapply)
       )
 
+    private def toPrefixesToRemove(text: Option[String]): Seq[String] = {
```
rdawson-hmrc, line 111: Not sure it matters, but this looks like it will produce a seq with a single item (empty string) for None and Some(""), shouldn't it produce an empty sequence instead?

PaulCDurham: Good question, I'll test that out.

PaulCDurham: Fixed and added a test to prove that an empty string results in an empty list

rdawson-hmrc (approved):

PR #298, opened 2024-08-12
HIPP-1415 deep search

rdawson-hmrc:


app/assets/javascripts/exploreApis.js
```diff
@@ -51,65 +38,166 @@ export function onPageShow() {
...
+            view.showFilterResultsPanel(model.resultCount, model.filteredCount);
+        } else {
+            view.hideFilterResultsPanel();
+        }
     }
 
-    view.onFiltersChanged(() => {
-        applyFilters();
+    filters.forEach(filter => filter.initialise(model.apis));
+    filters.forEach(filter=> filter.onChange(applyFiltersAndPagination));
```
victorarbuesmallada, line 116: not sure if this can be done this way:
```suggestion
    filters.forEach(filter => {
        filter.initialise(model.apis));
        filter.onChange(applyFiltersAndPagination);
    });
```

app/assets/javascripts/exploreApisResultPanels.js
```diff
@@ -0,0 +1,72 @@
...
+
+    let onClearFiltersHandler = noop;
+    elClearFilters.addEventListener('click', () => {
+        onClearFiltersHandler();
+    });
+
+    return {
+        show(showingCount, hiddenCount) {
+            elFilterResultsCount.textContent = showingCount;
+            elFilterResultsCountPlural.textContent = showingCount === 1 ? '' : 's';
```
victorarbuesmallada, line 59: do you reckon this is something we could add to `utils.js`?

app/views/ExploreApisView.scala.html
```diff
@@ -271,24 +265,59 @@ <h2 class="govuk-fieldset__heading">
...
+
+            <div id="apiResults" class="govuk-!-display-none">
+                <div id="searchResultsPanel" class="call-out-panel call-out-panel__small govuk-!-display-none">
+                    <p id="resultsSuccess" class="govuk-body information--general-tag govuk-!-margin-bottom-1">
+                        <i class="fa-solid fa-circle-info" aria-hidden="true"></i>
+                        <span class="call-out-type govuk-!-font-weight-regular">
+                            <span id="searchResultsShowing">Showing</span>
+                            <span id="searchResultsCount" class="govuk-!-font-weight-bold">0</span> result<span id="searchResultsCountPlural">s</span>
+                            for the search
+                            <span id="searchResultsTerm" class="govuk-!-font-weight-bold"></span>
```
victorarbuesmallada, line 284: is it worth having this coming from `messages`?

victorarbuesmallada (dismissed):
LGMT, small ignorable comments

app/views/ExploreApisView.scala.html
rdawson-hmrc: I try to put most text in messages but there are a few places where sentences get broken up into lots of little pieces including HTML (like here) with some bits shown/hidden by javascript and I think things just get very confusing if we try to put that into the messages file

victorarbuesmallada: makes sense :)

victorarbuesmallada (approved):

PR #311, opened 2024-08-20
HIPP-1459: Redirect api-hub requests to integration-hub

victorarbuesmallada:
There were some conflicts with the other routes so I went with the [health.Routes](https://github.com/hmrc/bootstrap-play/blob/main/bootstrap-health-play-30/src/main/scala/health/HealthRoutes.scala) approach

### DIFF
--- a/app/routes/OldRoutes.scala
+++ b/app/routes/OldRoutes.scala
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package routes
+
+import play.api.mvc._
+import play.api.routing.Router.Routes
+import play.api.routing.SimpleRouter
+import play.api.routing.sird._
+
+import javax.inject.Inject
+
+class OldRoutes @Inject()(
+                           val cc: MessagesControllerComponents
+                         ) extends SimpleRouter {
+  private val oldPath = "/api-hub"
+  private val newPath = "/integration-hub"
+
+  private object RedirectController extends ControllerHelpers {
+    private val Action = new ActionBuilder.IgnoringBody()(controllers.Execution.trampoline)
+
+    def redirect(to: String): Action[AnyContent] = Action {
+      Redirect(s"$newPath$to", statusCode = MOVED_PERMANENTLY)
+    }
+  }
+
+  override def routes: Routes = {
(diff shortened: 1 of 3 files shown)


conf/prod.routes
```diff
@@ -1,4 +1,4 @@
 # Add all the application routes to the app.routes file
-->         /api-hub                   app.Routes
 ->         /integration-hub           app.Routes
 ->         /                          health.Routes
+->         /                          routes.OldRoutes
```
PaulCDurham, line 4: This should probably be `/api-hub` as we don't want to consider redirecting anything else. That would make the routing logic easier too.

app/routes/OldRoutes.scala
```diff
@@ -0,0 +1,68 @@
...
+  private val oldPath = "/api-hub"
+  private val newPath = "/integration-hub"
+
+  private object RedirectController extends FrontendBaseController {
+
+    def onRedirect(path: String): Action[AnyContent] =
+      Action(_ => MovedPermanently(s"$newPath$path"))
+
+    override protected def controllerComponents: MessagesControllerComponents = cc
+  }
```
PaulCDurham, line 41: Have you seen this? https://www.playframework.com/documentation/3.0.x/ScalaRouting#The-Default-Controller

app/routes/OldRoutes.scala
```diff
@@ -0,0 +1,68 @@
...
+          parameterTypes = Seq(classOf[String]),
+          verb = "GET",
+          path = this.prefix + newPath,
+          comments = "",
+          modifiers = Seq.empty
+        )
+      ).call(RedirectController.onRedirect(newPath))
+  }
+}
+
```
PaulCDurham, line 68: There's a lot of code here. Compare it to this...
https://github.com/hmrc/oas-discovery-proxy/blob/main/app/uk/gov/hmrc/oasdiscoveryproxy/controllers/OasDiscoveryRouter.scala

I'm wondering if you can combine SimpleRouter and the Default controller and do all this in one or two lines of code.

PaulCDurham (commented):
Could this be simpler?

app/routes/OldRoutes.scala
victorarbuesmallada: that was a great idea but shame I couldn't used it because the `controllers` package gets shadowed by the project's [controller](https://github.com/hmrc/api-hub-frontend/tree/main/app/controllers) package

app/routes/OldRoutes.scala
```diff
@@ -0,0 +1,45 @@
...
+import javax.inject.Inject
+
+class OldRoutes @Inject()(
+                           val cc: MessagesControllerComponents
+                         ) extends SimpleRouter {
+  private val oldPath = "/api-hub"
+  private val newPath = "/integration-hub"
+
+  private object RedirectController extends ControllerHelpers {
```
victorarbuesmallada, line 32: I've rewritten this controller to be a simplified version of the [Default controller](https://github.com/playframework/playframework/blob/main/core/play/src/main/scala/play/api/controllers/Default.scala#L22).

victorarbuesmallada: that was waffle, I'm injecting it now

app/routes/OldRoutes.scala
victorarbuesmallada: that was not correct, using the `Default` controller now

PaulCDurham (approved):

PR #313, opened 2024-08-21
HIPP-1434: Fix some QA raised defects

victorarbuesmallada:
Two acceptance criteria points failed because:
- teams were not ordered by name
- couldn't change the owning team on deleted applications

### DIFF
--- a/app/controllers/application/UpdateApplicationTeamController.scala
+++ b/app/controllers/application/UpdateApplicationTeamController.scala
@@ -45,18 +45,19 @@ class UpdateApplicationTeamController @Inject()(
 
   private val form = formProvider()
 
-  def onPageLoad(applicationId: String): Action[AnyContent] = (identify andThen applicationAuth(applicationId)) async {
+  def onPageLoad(applicationId: String): Action[AnyContent] = (identify andThen applicationAuth(applicationId, includeDeleted = true)) async {
     implicit request => showView(OK, form)
   }
 
   private def showView(code: Int, form: Form[_])(implicit request: ApplicationRequest[_]): Future[Result] = {
     apiHubService.findTeams(None).map(teams => {
+      val sortedTeams = teams.sortBy(_.name.toLowerCase)
       val owningTeam = teams.find(team => request.application.teamId.contains(team.id))
-      Status(code)(view(form, request.application, owningTeam, teams, request.identifierRequest.user))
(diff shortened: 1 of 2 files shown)


app/controllers/application/ApplicationDetailsController.scala
```diff
@@ -38,7 +38,7 @@ class ApplicationDetailsController @Inject()(
   errorResultBuilder: ErrorResultBuilder
 )(implicit ec: ExecutionContext) extends FrontendBaseController with I18nSupport {
 
-  def onPageLoad(id: String): Action[AnyContent] = (identify andThen applicationAuth(id, enrich = true)).async {
+  def onPageLoad(id: String): Action[AnyContent] = (identify andThen applicationAuth(id, enrich = true, includeDeleted = true)).async {
```
PaulCDurham: We don't want to do this. It would make a deleted application visible to all users. Deleted applications will only be accessed via the Team migration page anyway.

app/controllers/application/UpdateApplicationTeamController.scala
```diff
@@ -45,14 +45,15 @@ class UpdateApplicationTeamController @Inject()(
...
   private val form = formProvider()
 
-  def onPageLoad(applicationId: String): Action[AnyContent] = (identify andThen applicationAuth(applicationId)) async {
+  def onPageLoad(applicationId: String): Action[AnyContent] = (identify andThen applicationAuth(applicationId, includeDeleted = true)) async {
     implicit request => showView(OK, form)
   }
 
   private def showView(code: Int, form: Form[_])(implicit request: ApplicationRequest[_]): Future[Result] = {
     apiHubService.findTeams(None).map(teams => {
+      val sortedTeams = teams.sortBy(_.name)
```
PaulCDurham, line 54: Use `_.name.toLowerCase` for a case-insensitive sort. This is a standard thing we do in the service.

app/controllers/application/ApplicationDetailsController.scala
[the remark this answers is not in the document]
victorarbuesmallada: that makes sense, I'll leave it on only on the screen that updates the application owning team

PaulCDurham (changes requested):
Could you check the unit tests for UpdateApplicationTeamController? I note you had to change unit tests for ApplicationDetailsController but not UpdateApplicationTeamController.

PR #316, opened 2024-08-23
HIPP-1434: Sort teams in an case insensitive way

victorarbuesmallada:


app/views/application/UpdateApplicationTeamView.scala.html
```diff
@@ -101,7 +101,7 @@ <h3 class="govuk-heading-s">@messages("application.update.team.selectNew")</h3>
 
                 <select id="owningTeam" name="owningTeam">
                     <option value="">Select an owning team</option>
-                    @for(team <- teams.sortBy(_.name)){
+                    @for(team <- teams.sortBy(_.name.toLowerCase)){
```
rdawson-hmrc, line 104: could we add a unit tests for this in the controller?

app/views/application/UpdateApplicationTeamView.scala.html
```diff
@@ -101,7 +101,7 @@ <h3 class="govuk-heading-s">@messages("application.update.team.selectNew")</h3>
 
                 <select id="owningTeam" name="owningTeam">
                     <option value="">Select an owning team</option>
-                    @for(team <- teams.sortBy(_.name)){
```
victorarbuesmallada, line 104: The controller was already sorting it the right way, this bit was messing it up

rdawson-hmrc (approved):

PR #323, opened 2024-08-29
HIPP-1468 - fixing broken acceptance tests

rdawson-hmrc:


victorarbuesmallada (dismissed):

app/views/components/ApplicationDetailsTable.scala.html
```diff
@@ -41,43 +41,40 @@
...
-                content = Text(ViewUtils.formatLocalDateTimeContainingUtc(application.created)),
-                classes = "utcDateLong"
-            ),
-        ),
-        SummaryListRow(
-            key = Key(
-                content = Text(messages("applicationDetailsTable.id"))
-            ),
-            value = Value(
-                content = application.id
```
victorarbuesmallada, line 76: I wonder if we could leave it as it was and have something like this here
```
content = Html(s"""<span id="applicationId">${application.id}</span""")
```
I'm fine with both options in any case

rdawson-hmrc: Yeah I thought about that but don't really like building html in strings it means the IDE doesn't give you any help when the syntax is wrong.

victorarbuesmallada: makes sense

victorarbuesmallada (approved):

PR #324, opened 2024-08-29
HIPP-1469: Display access request details on deleted applications

victorarbuesmallada:


app/views/admin/AccessRequestView.scala.html
```diff
@@ -52,6 +53,14 @@ <h2 class="govuk-heading-m">
             </dt>
             <dd class="govuk-summary-list__value">
                 @application.name
+                @if(application.isDeleted){
```
PaulCDurham, line 56: Twirl has a better way of handling optionals.
`@for(deleted <- application.deleted) { etc }`
Then you don't have to worry about mapping and orElse stuff. The whole code block is a more elegant map.

PaulCDurham (approved):

PR #330, opened 2024-08-30
HIPP-1477: UserModel changes

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1477

### DIFF
--- a/test/controllers/actions/StrideAuthenticatorSpec.scala
+++ b/test/controllers/actions/StrideAuthenticatorSpec.scala
@@ -17,19 +17,19 @@
 package controllers.actions
 
 import controllers.actions.StrideAuthenticator.{API_HUB_APPROVER_ROLE, API_HUB_PRIVILEGED_USER_ROLE, API_HUB_SUPPORT_ROLE, API_HUB_USER_ROLE}
-import controllers.actions.StrideAuthenticatorSpec._
+import controllers.actions.StrideAuthenticatorSpec.*
 import models.user.{Permissions, StrideUser, UserModel}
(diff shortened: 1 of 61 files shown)


test/controllers/actions/StrideAuthenticatorSpec.scala
```diff
@@ -120,15 +116,34 @@ class StrideAuthenticatorSpec extends AsyncFreeSpec with Matchers with MockitoSu
...
+
+      when(authConnector.authorise(eqTo(userPredicate), eqTo(userRetrieval))(any(), any()))
+        .thenReturn(Future.successful(retrievalsForUserWithoutEmail(user)))
+
+      strideAuthenticator.authenticate()(FakeRequest()).map {
+        result =>
+          result mustBe UserMissingEmail(user.userId, StrideUser)
+      }
+    }
+
```
victorarbuesmallada, line 138: do we need a test for when the [provider id is missing](https://github.com/hmrc/api-hub-frontend/pull/330/files?diff=unified&w=1#diff-5522eb046a7ed7f1694df8a76abbcaf0de6230611b702d12d6718a701f530826R62)?

victorarbuesmallada (dismissed):

test/controllers/actions/StrideAuthenticatorSpec.scala
PaulCDurham: I'm not sure this can ever happen. The Provider Id is aka the PID and the Stride login stub says you should always enter one. However if this does happen we want to know the behaviour so I'll add a test.

victorarbuesmallada (approved):

PR #338, opened 2024-09-04
HIPP-1482: Add filter to manage applications page

victorarbuesmallada:


app/views/admin/ManageApplicationsView.scala.html
```diff
@@ -44,12 +47,14 @@
     @views.html.templates.SideNav(messages("manageApplications.admin"), AdminNavItems(user, AdminSideNavPages.ManageApplicationsPage)) {
 
         <h1 class="govuk-heading-m govuk-!-margin-bottom-7">
-            @messages("manageApplications.heading", applications.size)
+            @messages("manageApplications.heading") (<span id="appCount">@{applications.size}</span>)
```
llmikeyj, line 50: Should this be

`@Html(messages("manageApplications.heading", <span id="appCount">@{applications.size}</span>))
`
?

llmikeyj: I think actually that both approaches are equivalent, but if you make the change I describe above you'll need to revert the change to the messages.en file

llmikeyj (approved):

PR #344, opened 2024-09-09
HIPP-1492: Add remove endpoints functionality

victorarbuesmallada:


test/controllers/AddAnApiSelectEndpointsControllerSpec.scala
```diff
@@ -140,44 +140,6 @@ class AddAnApiSelectEndpointsControllerSpec extends SpecBase with MockitoSugar w
       }
     }
 
-    "must add further endpoints to an existing answer (add endpoints journey)" in {
```
victorarbuesmallada, line 143: this test is not needed now that we can take all the endpoints from the form [here](https://github.com/hmrc/api-hub-frontend/pull/344/files#diff-10b0e1c9e03535dd312a0bfe2ff0cd48e72546c5f2aea3b7cac22ccdddcf36b3L170)

rdawson-hmrc (approved):

PR #345, opened 2024-09-11
HIPP-1502: Remove team from api or application

llmikeyj:


app/services/ApiHubService.scala
```diff
@@ -180,11 +180,15 @@ class ApiHubService @Inject()(
...
+    teamId match
+      case "unassigned" => applicationsConnector.removeApiTeam(apiId)
+      case _ => applicationsConnector.updateApiTeam(apiId, teamId)
   }
 
-  def changeOwningTeam(applicationId: String, teamId: String)(implicit hc: HeaderCarrier): Future[Option[Unit]] = {
-    applicationsConnector.changeOwningTeam(applicationId, teamId)
+  def updateApplicationTeam(applicationId: String, teamId: String)(implicit hc: HeaderCarrier): Future[Option[Unit]] = {
+    teamId match
+      case "unassigned" => applicationsConnector.removeApplicationTeam(applicationId)
```
PaulCDurham, line 190: I don't think we should be passing the UI drop down value around. The form should map to and from Option[String] and we should then pass that between layers. Have a look at ChangeTeamNameFormProvider for an example of the transform method.

PaulCDurham (approved):

PR #358, opened 2024-09-18
wip

victorarbuesmallada:


victorarbuesmallada:
Now it's just the link pointing to the Google Feedback page.

app/views/templates/Layout.scala.html
```diff
@@ -94,7 +94,23 @@
...
 }
 
+
+@feedbackTarget = @{
+    if (user.exists(_.userType != StrideUser)) {
+        "_blank"
+    } else {
+        "_self"
+    }
+}
```
PaulCDurham, line 104: Does this always work. You can sign-in with LDAP on a Stride machine. Is it more to do with the browser/OS than the user maybe?

victorarbuesmallada: I don't know, I'm just going with what the ACs seem to require
```
Scenario: Navigate to feedback form
Given I am on any page of the Hub
When I select the feedback link
Then I am navigated to the feedback form
And this opens in a new tab

Scenario: Stride user
Given I am using a Stride machine
When I select the feedback link
Then the feedback form opens
```

I'll get more context from Whitney

victorarbuesmallada: I've just changed that to use the `user-agent` instead.
I'm not on windows but got the testing values from this [website](https://www.useragentstring.com/pages/useragentstring.php)

conf/messages.en
```diff
@@ -760,3 +760,5 @@ produceApiEnterOas.hint = We''ve pre-populated the editor below with the specifi
 produceApiEnterOas.checkYourAnswersLabel = OpenAPI Specification
 produceApiEnterOas.error.required = Enter a valid OpenAPI Specification for your API
 produceApiEnterOas.change.hidden = OpenAPI Specification
+
+feedback.button = Feedback
```
rdawson-hmrc, line 764: we've been a bit inconsistent about this but I think we're supposed to prefix site-wide text with 'site.' (lots of examples at the top of the file)

app/assets/stylesheets/application.scss
```diff
@@ -43,6 +43,8 @@ $secondary-success-color: govuk-colour("white");
 $secondary-success-background: govuk-colour("dark-grey");
 $active-link-color: govuk-colour("dark-blue");
 
+$font-family: "GDS Transport", arial, sans-serif;
```
rdawson-hmrc, line 46: nice, we should do more of this

app/views/ViewUtils.scala
```diff
@@ -16,16 +16,19 @@
...
+import play.api.mvc.RequestHeader
 import play.twirl.api.Html
 
-import java.time.{Instant, LocalDateTime, ZoneOffset}
 import java.time.format.DateTimeFormatter
+import java.time.{Instant, LocalDateTime, ZoneOffset}
 
 object ViewUtils {
 
+  private val userAgentValuesStrideBrowsers = Seq("edge", "trident", "msie")
```
rdawson-hmrc, line 30: we can remove this now I think

rdawson-hmrc (approved):

PR #360, opened 2024-09-19
HIPP-1532: View HoD and Domain config

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1532

### DIFF
--- a/conf/messages.en
+++ b/conf/messages.en
@@ -778,3 +778,12 @@ produceApiBeforeYouStart.relatedContent.producingApis = Producing APIs
 produceApiBeforeYouStart.relatedContent.consumingApis = Consuming APIs
 
 relatedContent.header = Related content
+
+configuration.title = API configuration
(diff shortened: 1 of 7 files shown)


conf/messages.en
```diff
@@ -778,3 +778,12 @@ produceApiBeforeYouStart.relatedContent.producingApis = Producing APIs
 produceApiBeforeYouStart.relatedContent.consumingApis = Consuming APIs
 
 relatedContent.header = Related content
+
+configuration.title = API configuration
+configuration.domains.heading = Domains data
+configuration.domains.code = Code
+configuration.domains.description = Description
```
victorarbuesmallada, line 785: I wonder if we could have this as `site.description` instead, so we could reuse it (we are already using it in `apiDetails.endpoints.description`)

victorarbuesmallada (approved):

PR #392, opened 2024-10-08
HIPP-1549: Cancel prod access request navigation

llmikeyj:


PaulCDurham (changes requested):
Can you change the navigation to match AC's? The access request Id vs API Id debate is more subjective and up to you. I think access request Id is more future proof.

PaulCDurham (approved):

PR #395, opened 2024-10-09
HIPP-1549: Banner fixes

llmikeyj:


app/views/application/ApplicationAccessRequestsView.scala.html
```diff
@@ -22,21 +22,29 @@
...
     accessRequestPanel: AccessRequestPanel,
-    applicationDetailsTable: ApplicationDetailsTable
+    applicationDetailsTable: ApplicationDetailsTable,
+    iconsLink: IconsLink
 )
 
 @(application: Application, accessRequests: Seq[AccessRequest], user: UserModel)(implicit request: Request[?], messages: Messages)
 
+@pendingAccessRequestsCount = @{
+    accessRequests.count(_.status == Pending)
```
victorarbuesmallada, line 39: not sure if you would prefer to use exists instead
```suggestion
    accessRequests.exists(_.status == Pending)
```

app/controllers/helpers/ApplicationApiBuilder.scala
```diff
@@ -31,11 +31,10 @@ class ApplicationApiBuilder @Inject()(
   apiHubService: ApiHubService
 )(implicit ec: ExecutionContext) extends FrontendHeaderCarrierProvider {
 
-  def build(application: Application)(implicit request: Request[?]): Future[Seq[ApplicationApi]] = {
+  def build(application: Application)(implicit request: Request[?]): Future[(Seq[ApplicationApi], Int)] = {
```
PaulCDurham, line 34: You're kind of binding a generic component to specific UI usage. If you have to do it this way at least use a case class rather than a tuple. The method return of `Future[(Seq[ApplicationApi], Int)]` gives no indication of what the `Int` is.

Maybe add a utility method to the companion object for ApplicationApi that can give you just the Int? That would be cleaner.

victorarbuesmallada (approved):

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #397, opened 2024-10-11
HIPP-1252 - adding id to link

rdawson-hmrc:


edgarjimenez (approved):

PaulCDurham (changes requested):
I'm not sure that's sufficient to allow the tests to link back to the application. It should probably look like the register application success link.

PR #413, opened 2024-10-18
Preformat OAS global error content

victorarbuesmallada:


PaulCDurham (dismissed):

app/viewmodels/govuk/ErrorSummaryFluency.scala
```diff
@@ -29,14 +29,19 @@ trait ErrorSummaryFluency {
...
              )(implicit messages: Messages): ErrorSummary = {
 
       val errors = form.errors.map {
         error =>
+          val errormessage = messages(error.message, error.args*)
           ErrorLink(
             href    = Some(s"#${errorLinkOverrides.getOrElse(error.key, error.key)}"),
-            content = Text(messages(error.message, error.args*))
+            content = if (preFormat)
+                HtmlContent(s"<pre>$errormessage</pre>")
```
rdawson-hmrc, line 42: this would let apim perform a xss attack on the hub, but if they are doing that we will have bigger problems so it's probably fine:
errorMessage = '</pre><script>do something evil</script><pre>'

rdawson-hmrc (approved):

app/viewmodels/govuk/ErrorSummaryFluency.scala
victorarbuesmallada: that's a fair point, I've just added a bit to escape the html on that error message

PR #414, opened 2024-10-18
HIPP-1611 list apis in prod stat

rdawson-hmrc:


app/assets/javascripts/hubStatistics.js
```diff
@@ -1,17 +1,30 @@
 import {noop} from "./utils.js";
 
-function buildStatDisplay(el) {
+function buildCopyLink(valueToCopy, elContainer) {
```
victorarbuesmallada, line 3: do you reckon we could move this to a utils file? or else we can do that whenever we need to copy something in future tickets

victorarbuesmallada (approved):

PR #422, opened 2024-10-24
HIPP-1615: Adopt new APIM deployment statuses

victorarbuesmallada:


test/controllers/myapis/ApiUsageControllerSpec.scala
```diff
@@ -21,11 +21,12 @@ import controllers.actions.{FakeSupporter, FakeUser}
 import controllers.routes
 import generators.ApiDetailGenerators
 import models.api.ApiDeploymentStatuses
+import models.api.ApiDeploymentStatus.*
 import models.application.ApplicationLenses.*
-import models.application.{Api, Application, Creator, Deleted, TeamMember}
+import models.application.{Api, Application, Creator, Deleted, Primary, Secondary, TeamMember}
 import models.team.Team
 import models.user.UserModel
-import org.mockito.ArgumentMatchers.{any, eq as eqTo}
```
victorarbuesmallada, line 28: IntelliJ messes up with `as` whenever it needs to deal with imports, but it's fine with `=>`

app/models/api/ApiDeploymentStatuses.scala
```diff
@@ -16,12 +16,59 @@
...
+object ApiDeploymentStatus {
+
+  case class Deployed(override val environmentName: EnvironmentName, version: String) extends ApiDeploymentStatus
+
+  case class NotDeployed(override val environmentName: EnvironmentName) extends ApiDeploymentStatus
+
+  case class Unknown(override val environmentName: EnvironmentName) extends ApiDeploymentStatus
+
+  implicit val readDeploymentStatus: Reads[ApiDeploymentStatus] =
+    ((__ \ "_type").read[String]
```
victorarbuesmallada, line 38: the `_type`  has a different package in the backend and needed to do this to get around that.

app/controllers/myapis/MyApiDetailsController.scala
```diff
@@ -41,13 +42,9 @@ class MyApiDetailsController @Inject()(
...
     implicit request => for {
       deploymentStatuses <- apiHubService.getApiDeploymentStatuses(request.apiDetails.publisherReference)
-      team <- deploymentStatuses.flatMap(_ => request.apiDetails.teamId.map(apiHubService.findTeamById)).getOrElse(Future.successful(None))
-    } yield deploymentStatuses match {
-      case Some(deploymentStatuses) =>
-        Ok(view(request.apiDetails, deploymentStatuses, request.identifierRequest.user, config.supportEmailAddress, team.map(_.name)))
-      case None =>
-        errorResultBuilder.internalServerError(s"Unable to retrieve deployment statuses for API ${request.apiDetails.publisherReference}")
-    }
+      team <- request.apiDetails.teamId.flatTraverse(apiHubService.findTeamById)
```
PaulCDurham, line 45: Do we need Cats for this? Remember not everybody knows it. If it gives us something useful like EitherT then Ok but I don't see what value this method is adding.

PaulCDurham: Also, this is ignoring the case where the team is not found. Maybe look at the very similar ApiDetailsController and how that handles teams.

Maybe abstract a common way of handling an optional team Id to optional text and add that to the Fetching trait.

PaulCDurham (changes requested):

PaulCDurham (approved):

PR #428, opened 2024-10-29
HIPP-1618: Fallback to  when content is not provided on requestBody

victorarbuesmallada:
There are cases in which requests are not provided within a `content` block, which is still valid as we can see in this [example](https://admin.qa.tax.service.gov.uk/integration-hub/apis/oas/ae7cfc2c-f4d2-47e9-957b-76f44be9ad15).
In those cases, the requests bodies are provided as `$ref` values.

These changes add the fallback for when the `content` is not there but `$ref` is.

### DIFF
--- a/app/models/curl/OpenApiDoc.scala
+++ b/app/models/curl/OpenApiDoc.scala
@@ -19,6 +19,7 @@ package models.curl
 import com.fasterxml.jackson.databind.module.SimpleModule
 import io.swagger.oas.inflector.examples.ExampleBuilder
 import io.swagger.oas.inflector.processors.JsonNodeExampleSerializer
+import io.swagger.v3.oas.models.media.{Content, MediaType, Schema}
 import io.swagger.v3.oas.models.{Components, OpenAPI, Operation, PathItem, Paths}
 import io.swagger.v3.oas.models.servers.Server
 import io.swagger.v3.parser.OpenAPIV3Parser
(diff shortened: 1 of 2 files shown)


victorarbuesmallada:
test this please

victorarbuesmallada:
test this please

app/models/curl/OpenApiDoc.scala
```diff
@@ -90,6 +92,13 @@ class OpenApiDoc(openApi: OpenAPI) {
         .map(param => param.getName -> (if (includeNameInValue) missingValue(param.getName, param.getSchema.getType) else missingValue(param.getSchema.getType)))
         .toMap
     }
+
+    private def contentFromRef(maybe$ref: Option[String]): Option[Content] =
+      maybe$ref.map { $ref =>
+        val schema = Schema().$ref($ref)
+        val mediaType = MediaType().schema(schema)
+        Content().addMediaType("application/json", mediaType)
```
rdawson-hmrc, line 100: can we assume json here? shouldn't the first argument be derived from the contents of the ref somehow?

victorarbuesmallada: we could, the `$ref` attribute is just a string, but we could parse that and get the content type out of that

victorarbuesmallada: it was actually simpler, the `Schema` class already contained a content type field.

rdawson-hmrc (approved):

PR #434, opened 2024-11-05
HIPP-1628: Diagnostic scope view

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1628

### DIFF
--- a/app/connectors/ApplicationsConnector.scala
+++ b/app/connectors/ApplicationsConnector.scala
@@ -599,7 +599,7 @@ class ApplicationsConnector @Inject()(
       .execute[Seq[EgressGateway]]
   }
 
-  def fetchALlScopes(applicationId: String)(implicit hc: HeaderCarrier): Future[Option[Seq[CredentialScopes]]] = {
+  def fetchAllScopes(applicationId: String)(implicit hc: HeaderCarrier): Future[Option[Seq[CredentialScopes]]] = {
     httpClient.get(url"$applicationsBaseUrl/api-hub-applications/applications/$applicationId/all-scopes")
       .setHeader(ACCEPT -> JSON)
       .setHeader(AUTHORIZATION -> clientAuthToken)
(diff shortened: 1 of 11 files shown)


app/services/ApiHubService.scala
```diff
@@ -214,4 +214,8 @@ class ApiHubService @Inject()(
     applicationsConnector.listEgressGateways()
   }
 
+  def fetchAllScopes(applicationId: String)(implicit hc: HeaderCarrier): Future[Option[Seq[CredentialScopes]]] = {
+    applicationsConnector.fetchALlScopes(applicationId)
```
victorarbuesmallada, line 218:
```suggestion
    applicationsConnector.fetchAllScopes(applicationId)
```

test/viewmodels/application/ApplicationNavItemsSpec.scala
```diff
@@ -133,7 +141,9 @@ class ApplicationNavItemsSpec extends SpecBase with Matchers with TestHelpers wi
       }
     }
```
victorarbuesmallada, line 143: Should there be a test case for the scenario in which the `ManageTeamMembers` nav item is hidden when the team is not migrated:
https://github.com/hmrc/api-hub-frontend/pull/434/files#diff-fc66007bfa67e5e66f02aa4442334c9a89c7488eb2b6d8964244a780012c3ae5R63

victorarbuesmallada (approved):

test/viewmodels/application/ApplicationNavItemsSpec.scala
PaulCDurham: That nav item should not be hidden when the team is not migrated so no I don't think we need that test case.

app/connectors/ApplicationsConnector.scala
```diff
@@ -598,4 +598,17 @@ class ApplicationsConnector @Inject()(
       .setHeader(AUTHORIZATION -> clientAuthToken)
       .execute[Seq[EgressGateway]]
   }
+
+  def fetchALlScopes(applicationId: String)(implicit hc: HeaderCarrier): Future[Option[Seq[CredentialScopes]]] = {
```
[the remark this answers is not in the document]
PaulCDurham: Typo, I'll fix that

victorarbuesmallada (approved):

PR #437, opened 2024-11-06
HIPP-1595 - Deploy API - Producer

victorarbuesmallada:


app/controllers/myapis/produce/ProduceApiCheckYourAnswersController.scala
```diff
@@ -38,34 +45,142 @@ class ProduceApiCheckYourAnswersController @Inject()(
...
+        passthrough <- validatePassthrough(request.userAnswers, request.user.permissions.canSupport)
+        apiStatus <- validateApiStatus(request.userAnswers)
+        domainSubdomain <- validateDomainSubdomain(request.userAnswers)
+        hods <- validateHods(request.userAnswers)
+        egressPrefixes <- validateEgressPrefixes(request.userAnswers)
+    } yield DeploymentsRequest(
+      lineOfBusiness = "apim",
+      name = apiTitle,
+      description = shortDescription,
+      egress = egress.egress.getOrElse("unassigned"),
```
victorarbuesmallada, line 86: I've seen that there is some logic around the `unassigned` value on the [egress form](https://github.com/hmrc/api-hub-frontend/blob/main/app/forms/myapis/produce/ProduceApiChooseEgressFormProvider.scala#L37-L48), not sure if I should change that logic

app/services/ApiHubService.scala
```diff
@@ -218,4 +225,12 @@ class ApiHubService @Inject()(
     applicationsConnector.fetchAllScopes(applicationId)
   }
 
+  private[services] def formatAsKebabCase(text: String): String =
```
victorarbuesmallada, line 228: not sure if this should be here, on a utils class or if should be on the `api-hub-applications` side, happy to change it though

app/views/myapis/produce/ProduceApiCheckYourAnswersView.scala.html
```diff
@@ -22,18 +23,38 @@
     formHelper: FormWithCSRF,
     govukButton: GovukButton,
     govukSummaryList: GovukSummaryList,
+    govukErrorSummary: GovukErrorSummary,
 )
 
-@(summaryList: SummaryList, userModel: UserModel)(implicit request: Request[?], messages: Messages)
+@(summaryList: SummaryList, userModel: UserModel, maybeFailuresResponse: Option[FailuresResponse] = None)(implicit request: Request[?], messages: Messages)
+
+@errorList(failuresResponse: FailuresResponse) = @{
```
victorarbuesmallada, line 31: not sure if we want to display errors this way, it looks this way:
Screenshot from 2024-11-06 17-02-45

app/services/ApiHubService.scala
PaulCDurham: Put this in the controller? That instantiates the DeploymentsRequest.

app/views/myapis/produce/ProduceApiCheckYourAnswersView.scala.html
PaulCDurham: Maybe pre tags around pretty printed JSON like we did when validating the OAS.

victorarbuesmallada: Screenshot from 2024-11-07 15-36-56

PR #438, opened 2024-11-07
HIPP-1636: Hide 'Cancel request' button when the access request is no…

victorarbuesmallada:
…t on 'Pending' state and the user does not belong to the application team


llmikeyj (changes requested):
I think we need one more test case, which is that even if you are an approver, you cannot cancel the request when you are not a team member?

app/controllers/application/AccessRequestController.scala
```diff
@@ -50,19 +51,19 @@ class AccessRequestController @Inject()(
     implicit request =>
       (for {
         accessRequest <- EitherT(fetchAccessRequestOrNotFound(accessRequestId))
-        application <- EitherT(fetchApplicationOrNotFound(accessRequest.applicationId))
-      } yield buildView(accessRequest, application)).merge
+        result <- EitherT.right(
+          applicationAuth(accessRequest.applicationId, enrich = false)
+            .invokeBlock(request, implicit applicationRequest => Future.successful(buildView(accessRequest)))
```
victorarbuesmallada, line 56: this endpoint doesn't take an application id as parameter, so I needed to use the action builder like this

victorarbuesmallada:
> I think we need one more test case, which is that even if you are an approver, you cannot cancel the request when you are not a team member?

The logic behind this ticket has changed and is now about hiding the button only if the screen is on the *admin* side of the service.

PaulCDurham (approved):

PR #440, opened 2024-11-08
HIPP-1647: Create the skeleton journey for Update an API flow

victorarbuesmallada:


victorarbuesmallada:
I've added the second commit just in case we don't use the user answers to keep the id of the API to update.

PR #441, opened 2024-11-11
HIPP-1625: Fix scope disparity

PaulCDurham:
https://jira.tools.tax.service.gov.uk/browse/HIPP-1625

### DIFF
--- a/app/views/application/AllScopesView.scala.html
+++ b/app/views/application/AllScopesView.scala.html
@@ -1,0 +1,73 @@
+@*
+ * Copyright 2024 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


app/views/application/AllScopesView.scala.html
```diff
@@ -0,0 +1,73 @@
...
+            head = Some(Seq(
+              HeadCell(
+                content = Text("Environment")
+              ),
+              HeadCell(
+                content = Text("Client Id")
+              ),
+              HeadCell(
+                content = Text("Scopes")
+              )
```
victorarbuesmallada, line 61: could those be in the `messages.en` file? we have those in there for [different pages](https://github.com/hmrc/api-hub-frontend/blob/main/conf/messages.en#L226-L228), not sure if we can have them as `site.textKey`

victorarbuesmallada (approved):
LGTM, just some text that could go on the messages file

PR #442, opened 2024-11-11
HIPP-1640 - Add ipaas SRS roles to The Hub

victorarbuesmallada:


app/controllers/actions/StrideAuthenticator.scala
```diff
@@ -38,11 +39,9 @@ class StrideAuthenticator @Inject()(
...
   def authenticate()(implicit request: Request[?]): Future[UserAuthResult] = {
     authorised(
-      Enrolment(API_HUB_USER_ROLE) or
-        Enrolment(API_HUB_APPROVER_ROLE) or
-        Enrolment(API_HUB_SUPPORT_ROLE) or
-        Enrolment(API_HUB_PRIVILEGED_USER_ROLE) and
-        AuthProviders(PrivilegedApplication)
+      (approverRoles ++ privilegedRoles ++ supportRoles ++ userRoles)
+        .foldRight[Predicate](EmptyPredicate)(Enrolment(_) or _)
+          and AuthProviders(PrivilegedApplication)
```
PaulCDurham, line 44: This approach will put ipaas_live_admins into the list twice. Hopefully that doesn't cause an issue. Using sets might have been better.

app/controllers/actions/StrideAuthenticator.scala
```diff
@@ -87,4 +86,13 @@ object StrideAuthenticator {
   val API_HUB_SUPPORT_ROLE: String = "api_hub_support"
   val API_HUB_PRIVILEGED_USER_ROLE: String = "api_hub_privileged_user"
 
+  val IPAAS_LIVE_SERVICE: String = "ipaas_live_service"
+  val IPAAS_LIVE_ADMINS: String = "ipaas_live_admins"
+  val IPAAS_LIVE_SERVICE_SC: String = "ipaas_live_service_sc"
```
PaulCDurham, line 91: It would have been better to stick with the existing _ROLE naming convention for role names.

PaulCDurham (approved):

PR #464, opened 2024-11-22
HIPP-1671: Show the cancel request link on access requests that are p…

victorarbuesmallada:
…ending


test/controllers/application/AccessRequestControllerSpec.scala
```diff
@@ -78,6 +78,33 @@ class AccessRequestControllerSpec
       }
     }
 
+    "must not display the cancel request link if the access request is cancelled" in {
```
PaulCDurham, line 81: More correctly, we don't want to display the link when the status is not pending. Maybe change the forAll to test all non-pending status values with a single correct user (FakeUser should do)?

Change the test description to something like "must not display the cancel link for non-pending statuses"

PaulCDurham (approved):

PR #471, opened 2024-12-04
HIPP-1744: Add Deployment request error screen

victorarbuesmallada:


app/views/myapis/produce/ProduceApiDeploymentErrorView.scala.html
```diff
@@ -14,22 +14,91 @@
...
+    <div class="govuk-button-group">
         @govukButton(
-            ButtonViewModel("Return to API deployment page")
+            ButtonViewModel(messages("produceApiDeploymentError.return"))
+                .asLink(controllers.myapis.produce.routes.ProduceApiCheckYourAnswersController.onPageLoad().url)
         )
-    }
+
+        <a class="govuk-link govuk-link--no-visited-state govuk-!-margin-left-3" href="@routes.IndexController.onPageLoad.url">@messages("site.cancel")</a>
+    </div>
```
PaulCDurham, line 103: When we cancel it would be nice to clear the user answers collection. That way the user can no longer re-enter the journey.

app/controllers/myapis/produce/ProduceApiCheckYourAnswersController.scala
```diff
@@ -60,12 +62,14 @@ class ProduceApiCheckYourAnswersController @Inject()(
...
       validate(request).fold(
         call => Future.successful(Redirect(call)),
-        apiHubService.generateDeployment(_).map {
+        apiHubService.generateDeployment(_).flatMap {
           case response: SuccessfulDeploymentsResponse =>
-            Ok(successView(request.user, response))
-          case InvalidOasResponse(failure) => BadRequest(
-            view(SummaryListViewModel(summaryListRows(request.userAnswers)), request.user, Some(failure))
-          )
+            Future.successful(Ok(successView(request.user, response)))
```
PaulCDurham, line 67: We should clear the user answers before returning the success view to prevent the user from re-entering the journey and trying to deploy the same API.

app/controllers/myapis/produce/ProduceApiCheckYourAnswersController.scala
```diff
@@ -60,12 +62,14 @@ class ProduceApiCheckYourAnswersController @Inject()(
...
+          case InvalidOasResponse(failure) =>
+            for {
+              updatedAnswers <- Future.fromTry(request.userAnswers.set(ProduceApiDeploymentErrorPage, failure))
+              _ <- produceApiSessionRepository.set(updatedAnswers)
+            } yield Redirect(controllers.myapis.produce.routes.ProduceApiDeploymentErrorController.onPageLoad())
```
PaulCDurham, line 72: The old version returned the view. This is what we've done in all the other journeys. It just works better with back buttons etc.

app/controllers/myapis/produce/ProduceApiCheckYourAnswersController.scala
PaulCDurham: Still need to do this

app/services/ApiHubService.scala
```diff
@@ -92,7 +92,14 @@ class ApiHubService @Inject()(
...
   def generateDeployment(deploymentsRequest: DeploymentsRequest)(implicit hc: HeaderCarrier): Future[DeploymentsResponse] =
-    applicationsConnector.generateDeployment(deploymentsRequest)
+//    applicationsConnector.generateDeployment(deploymentsRequest)
+    Future.successful(InvalidOasResponse(
+      FailuresResponse(
+        "code",
+        "reason",
+        Some(Seq(models.deployment.Error("type", "message")))
+      )
+    ))
```
PaulCDurham, line 102: Stubbed out some code? Might be why the tests are failing.

PaulCDurham (changes requested):

app/controllers/myapis/produce/ProduceApiCheckYourAnswersController.scala
victorarbuesmallada: ah missed it, just committed the changes for that

PaulCDurham (approved):

PR #480, opened 2024-12-10
Hipp 1756 stop unnecessary enriching

llmikeyj:


test/controllers/application/CurlCommandControllerSpec.scala
```diff
@@ -67,6 +67,21 @@ class CurlCommandControllerSpec extends SpecBase with MockitoSugar with TestHelp
       }
     }
 
+    "must return OK and the correct JSON for a GET from a support user when the no credentials found" in {
```
victorarbuesmallada, line 70: just in case
```suggestion
    "must return OK and the correct JSON for a GET from a support user when no credentials are found" in {
```

victorarbuesmallada (approved):

PaulCDurham (approved):

PR #494, opened 2024-12-19
HIPP-1568: Add user model to error template view

victorarbuesmallada:


app/models/requests/BaseRequest.scala
```diff
@@ -0,0 +1,35 @@
...
+import play.api.i18n.MessagesProvider
+import play.api.mvc.{Request, RequestHeader, WrappedRequest}
+
+abstract class BaseRequest[T](request: Request[T]) extends WrappedRequest[T](request) {
+
+    def maybeUser: Option[UserModel]
+}
+
+object BaseRequest {
+    extension (request: RequestHeader)
```
victorarbuesmallada, line 29: I still needed this cause in some cases we can't pass a `BaseRequest` I think

PaulCDurham: If you used BaseRequest in place of Request in all the implicits in ErrorResultBuilder would you still need to do this? Assuming it still compiles Ok, but I think it should.

victorarbuesmallada:
test this please

app/models/requests/BaseRequest.scala
victorarbuesmallada: the issue is that I can't pass a `BaseRequest` [here](https://github.com/hmrc/api-hub-frontend/pull/494/files#diff-05404c3946fcb5849822e944516332e4dacf4a04ea3e36507dc6445167614d91L33), I can use the pattern match on that method though?

test/controllers/JourneyRecoveryControllerSpec.scala
```diff
@@ -35,14 +37,14 @@ class JourneyRecoveryControllerSpec extends SpecBase with HtmlValidation {
 
         running(application) {
           val continueUrl = RedirectUrl("/foo")
-          val request     = FakeRequest(GET, routes.JourneyRecoveryController.onPageLoad(Some(continueUrl)).url)
+          val request     = IdentifierRequest(FakeRequest(GET, routes.JourneyRecoveryController.onPageLoad(Some(continueUrl)).url), FakeUser)
```
PaulCDurham, line 40: We typically pass the user into SpecBase.applicationBuilder via the buildFixture() method.

rdawson-hmrc (approved):

PR #497, opened 2025-01-02
HIPP-1568: Render not found view with logged in user

victorarbuesmallada:
@saritaparigi found that when navigating to a non existing endpoint the default error handler for **not found** errors was not considering the logged in user.
This should fix that.


victorarbuesmallada:
test this please

rdawson-hmrc (approved):

PR #505, opened 2025-01-08
HIPP-1746: Remove EnvironmentName

llmikeyj:


victorarbuesmallada (dismissed):

conf/messages.en
```diff
@@ -42,6 +42,7 @@ site.signedIn = Signed in as {0} via {1}
 site.environment = Environment
 site.environment.dev = Development environment
 site.environment.test = Test environment
+site.environment.deployment = Deployment environment
```
PaulCDurham, line 45: Would we ever need this? The deployment environment should resolve to test and we'd use its name, ie "Test environment".

PaulCDurham (approved):

conf/messages.en
llmikeyj: No, forgot to tidy that up,

PaulCDurham (approved):

PR #512, opened 2025-01-13
HIPP-1817: Add egress to deploy responses and user answers

victorarbuesmallada:


app/controllers/myapis/update/UpdateApiStartController.scala
```diff
@@ -68,6 +68,8 @@ class UpdateApiStartController @Inject()(
           }
           updatedAnswers <- updateOptionalAnswer(updatedAnswers, UpdateApiDomainPage, domainSubdomain)
           updatedAnswers <- updateOptionalAnswer(updatedAnswers, UpdateApiHodPage, deploymentDetails.hods.map(_.toSet))
+          updatedAnswers <- updateAnswer(updatedAnswers, UpdateApiEgressAvailabilityPage, deploymentDetails.egress.exists(_ != "null-egress"))
+          updatedAnswers <- updateOptionalAnswer(updatedAnswers, UpdateApiEgressSelectionPage, deploymentDetails.egress.collect { case egress if egress != "null-egress" => egress })
```
rdawson-hmrc, line 72: Could we have this in just one place in the code, it appears in at least one other location at the moment

victorarbuesmallada: yeah I thought the same but couldn't think of where to place it :/

rdawson-hmrc: The idea might be to strongly-type the egress rather than just having it as a string, then we can include a method on the class for 'isNull' or something. If this is lots of work then I'd be happy with null-egress just being a constant somewhere

rdawson-hmrc (approved):

PR #534, opened 2025-01-24
HIPP-1850 - Improve promotion endpoints for V2

victorarbuesmallada:


app/controllers/myapis/SimpleApiPromotionController.scala
```diff
@@ -54,7 +54,12 @@ class SimpleApiPromotionController @Inject()(
     implicit request =>
       apiHubService.getApiDetail(id).flatMap {
         case Some(apiDetail) =>
-          applicationsConnector.promoteToProduction(apiDetail.publisherReference).map {
+          apiHubService.promoteAPI(
+            apiDetail.publisherReference,
+            hipEnvironments.deploymentHipEnvironment,
+            hipEnvironments.productionHipEnvironment,
+            DeploymentDetails.egressFallback
```
victorarbuesmallada, line 61: not sure if the new fields need to be passed from the frontend, my guess is that this screen will be replaced by the new promote one so it probably doesn't matter?

PaulCDurham: You could probably just remove the controller and the reference to it from the API details page.

PaulCDurham (approved):

PR #590, opened 2025-03-03
Hipp 1982 which apis pre production

llmikeyj:


app/controllers/application/accessrequest/RequestProductionAccessSelectApisController.scala
```diff
@@ -98,3 +107,16 @@ class RequestProductionAccessSelectApisController @Inject()(
         result.tupled(partitionApplicationApis(apis))
       case _ => Future.successful(Redirect(routes.JourneyRecoveryController.onPageLoad()))
     }
+
+  private def validateEnvironmentId(userAnswers: UserAnswers): Either[Call, HipEnvironment] = {
```
PaulCDurham, line 111: Why? This validation was done in the start controller. The user can't change the value. Just run with the value in user answers else journey recovery.

PaulCDurham (commented):
Slightly concerned that no ticket has responsibility for setting RequestProductionAccessApisPage based on environment. I guess we're hoping HIPP-1970 does that but it's not explicitly required their.

PR #599, opened 2025-03-07
HIPP-1983: Api environment access status

llmikeyj:


app/viewmodels/application/ApplicationApi.scala
```diff
@@ -116,6 +116,9 @@ case class ApplicationApi(
   def needsAccessRequest(hipEnvironment: HipEnvironment): Boolean = !hasPendingAccessRequest(hipEnvironment) && endpoints.exists(_.accessFor(hipEnvironment) == Inaccessible)
   def hasPendingAccessRequest(hipEnvironment: HipEnvironment): Boolean = pendingAccessRequests.exists(_.environmentId.contains(hipEnvironment.id))
   def isAccessibleInEnvironment(hipEnvironment: HipEnvironment): Boolean = availableEndpoints(hipEnvironment) > 0
+  def allEndpointsHavePendingAccessRequests(hipEnvironment: HipEnvironment): Boolean = {
```
rdawson-hmrc, line 119: is this used anywhere?

app/views/application/EnvironmentsViewApiTab.scala.html
```diff
@@ -78,11 +78,34 @@
...
-            <a class="govuk-link" href="@controllers.application.cancelaccessrequest.routes.CancelAccessRequestStartController.startJourney(viewModel.applicationId)">Cancel production access requests</a>
+            <a class="govuk-link" href="@controllers.application.cancelaccessrequest.routes.CancelAccessRequestStartController.startJourney(viewModel.applicationId)">Cancel access requests</a>
         </p>
     </div>
 }
 
+@apiStatus(applicationApi: ApplicationApi) = @{
+        (applicationApi.isAccessibleInEnvironment(viewModel.hipEnvironment), applicationApi.needsAccessRequest(viewModel.hipEnvironment)) match {
+            case (false, _) => "No access"
+            case (true, true) => "Partial access"
```
rdawson-hmrc, line 89: is this right? if they have access to some endpoints and have pending access requests for the others then that seems like it should qualify as 'partial access' but I think this will show 'full access'

app/viewmodels/application/ApplicationApi.scala
llmikeyj: Nope, thanks

app/views/application/EnvironmentsViewApiTab.scala.html
llmikeyj: Good spot

rdawson-hmrc (approved):

PR #613, opened 2025-03-26
Hipp 2055 better handling of deleted api

llmikeyj:


app/views/application/EnvironmentsViewApiTab.scala.html
```diff
@@ -105,7 +110,7 @@
                 @messages("applicationApis.apiMissing")
             </p>
             <p class="govuk-body">
-                <a class="govuk-link govuk-link--no-visited-state" href="@controllers.application.routes.RemoveApiController.onPageLoad(viewModel.applicationId, applicationApi.apiId, viewModel.hipEnvironment.id)" data-remove-api-api-id="@applicationApi.apiId">
+                <a class="govuk-link govuk-link--no-visited-state" href="@controllers.application.routes.RemoveApiController.onPageLoad(viewModel.applicationId, applicationApi.apiId)" data-remove-api-api-id="@applicationApi.apiId">
```
PaulCDurham, line 113: I think we need to drop Remove API from this page. Otherwise we need to return to this page once the API is removed. Which means adding environment Id back into all the endpoints you just removed it from. It's not in the prototype anymore.

llmikeyj: Yes, good call

